PR #23, opened 2019-06-20
added new mandate item about orchestrating the ecosystem (IT Strategy…

remyb2canada:
Added as a mandate item (our authority)

- Orchestrate the IT ecosystem by faclitating the execution of strategies, oversighting their executions, lead the changes identified in the strategies

This tells that we don't just produce paper, we execute


gabrielcossette:
Good idea, I agree! @jaysonmc?

PR #30, opened 2019-06-28
Proposing 8 KPIs to measure our strategies against

remyb2canada:
For team review, please comment

8 KPIs with which to measure our strategies against. The intent is to get buy-in from Denis and CIO. Those KPIs can be used to tell the story as to why strategy X is needed and why changes A, B, C are required.

KPIs will mean we will need to collect data. Data that is managed by other teams. So if we agree with those KPIs, it means we agree to establish relationships with other teams to collect such data.


Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+`ratio = service requests / service personnel`
+
+## 4. **Percentage of IITB services made available as SaaS**
+
+IITB services currently offered, as catalogued in the ITSM Repository, will be offered as SaaS for IITB consumers. This transformation to SaaS provisioning and maintenance will be measured by the rate at which those services are changed to self-served ones. The measurement will look at [IITB's Service Catalogue](http://iservice.prv/eng/service_catalogues/imit_services.shtml), which is being modernized for 2020, as reference point and monitor their transformation to SaaS.
+
+`percentage = SaaS / total ITSM services`
+
+## 5. **Availability uptime of ESDC core systems**
```
gcharest: That's a good KPI but I'm wondering if we can use it since we're not really in control of anything? For example, we could measure this current KPI and measure them IF our strategies are put in place.

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+Moving IITB into a service-oriented organization will require IITB to manage the complexity and growth of its services pro-actively. The measurement of availability is driven by time loss whereas the measurement of reliability is driven by the frequency and impact of failures. This indicator focuses on availability as it's what clients of the service are concerned with. The ***service's*** maximum allowable downtime, as agreed by the BCP Business Impact Analysis exercise, will be used to assess IITB's service availability. The Core systems will be the ones flagged as *high business value* within [ESDC's Application Portfolio Management (APM) program](http://dialogue/grp/PR6303013/APM_Program/default.aspx).
+
+`service availability = (total elapsed time - sum of downtime) / total elapsed time`
+
+*Note:* Services are never expected to be up 100% of the time. [See here](http://www.gcpedia.gc.ca/wiki/OCIO_Application_Portfolio_Management/Application_Portfolio_Management_User_Guide#Application_Mission_Criticality_and_Critical_Services) for more details.  
+
+## 6. **Increase rate of project intake by IITB**
```
gcharest: I don't know about project intake or even project completion/success rate. It may be out of our control as well and if we're successful, might accelerate or decelerate? It may turn into a product management culture instead and as such there would be less projects, more continuous development/improvement?

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+IITB's project intake capacity will be measured by how much it improves, fostering a continuous improvement measurement over a hard target. The measurement will be on the rate at which intake capacity increases. The PMO's [Active Project List](http://dialogue/grp/ippst/Approved%20Projects%20Lists%20%28APL%29/Forms/AllItems.aspx?RootFolder=%2Fgrp%2Fippst%2FApproved%20Projects%20Lists%20%28APL%29%2FFY%202019%2D2020&FolderCTID=0x012000D8E2E348AB0A034886495AE4BED12E07&View=%7b2C779E42-EB28-444B-AACD-39CC61DDBCE2%7d), for Major/Minor/Small ESDC projects, as well as the PMO's [IITB Lite projects](http://dialogue/grp/ITPM-GPTI/Lists/IITBPMP%20Small%20Projects/AllItems.aspx) will be used as reference points to monitor intake increase.
+
+`rate of intake = (Projects Year_n - Projects Year_n-1) / Projects Year_n-1`
+
+## 7. **Reduction in Cyber-incidents**
```
gcharest: This I really like. Should also be an increase in resilience testing somehwere in there: part of SOP, we throw a wrench in the system (in production!) and teams are able to solve the issues faster and faster.

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+This indicator will monitor the number of cyber-incidents across ESDC and measure the rate at which they decrease.
+
+`rate = (Incidents Year_n - Incidents Year_n-1) / Incidents Year_n-1`
+
+## 8. **Client Satisfaction of IITB Services**
```
gcharest: Well, that one is subjective but if we can move the needle to the right, that will be good!

gcharest:
@jaysonmc @gabrielcossette @arushisharma19 @smellems I encourage you to review in this thread to see if we want to make changes, adapt or reject some of the proposed KPIs.

Thanks!

gcharest:
@remyb2canada Great work putting those together! :grinning:

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+- Be efficient in managing its operations
+- Be connected with IM/IT partners and suppliers
+
+## 1. **Percentage of ESDC Services consumable by APIs**
+
+ESDC services provided to the public or to other departments will need to be delivered digitally. This means via APIs. This transformation to digital will be measured by the rate at which analogue services are changed to digital ones.
+
+The indicator will be measured by using [ESDC's Service Catalogue](https://www.canada.ca/en/employment-social-development/corporate/portfolio/service-canada/programs.html) as reference point and monitor the rate at which they are transformed to digital
+
+`percentage = digital services / total services`
```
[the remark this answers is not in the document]
remyb2canada: From an IT perspective, yes
The TB Digital Policy interpretation directive is "Every service must be consumable by an API, every API must be hosted on the GC API Store"

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+## 2. **Rate of decrease in processing time of ESDC Services**
```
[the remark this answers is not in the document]
remyb2canada: 1. The increase in personnel costs scenario is dealt with in 3. Reduction in toil

2. The increase in costs (large system deployment, expensive licences) is not covered, true. I see this as a financial matter that is part of the equation when designing services. The increase use of technology would probably also mean an increase ops costs for IITB due to the additional technology footprint across ESDC. Instead of having an indicator about op costs, I'd say 6. Increase rate of project intake by IITB be the indicator. This should counter large projects because if IITB can increase its rate of project intake, they'd have to be completed faster. Large projects would not.

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+Toil is the kind of work tied to running a production service that tends to be manual, repetitive, automatable, tactical devoid of enduring value, and that scales linearly as a service grows. Processing  time of ESDC services may decrease by increasing manual labour but that would defeat the purpose of digital.
+
+The indicator will be measured two ways:
+
+1. For toil in IITB Services, using ??? and measure the ratio of X over Y. 
+
+`ratio = X / Y`
+
+2. For toil in ESDC Services, using the [ESDC's Service Catalogue](https://www.canada.ca/en/employment-social-development/corporate/portfolio/service-canada/programs.html) and measure the ratio of service requests to amount of personnel.
```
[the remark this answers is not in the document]
remyb2canada: I'll reword the statement. I agree with what you're saying. The amount of personnel meant as support staff for the service. There will _always_ be personnel within the service delivery, whether it be to deal with exceptions or complex requests. That doesn't mean services cannot be automated from beginning to end.

Strategies/IITB_KPIs.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. For toil in IITB Services, using ??? and measure the ratio of X over Y. 
+
+`ratio = X / Y`
+
+2. For toil in ESDC Services, using the [ESDC's Service Catalogue](https://www.canada.ca/en/employment-social-development/corporate/portfolio/service-canada/programs.html) and measure the ratio of service requests to amount of personnel. 
+
+`ratio = service requests / service personnel`
+
+## 4. **Percentage of IITB services made available as SaaS**
```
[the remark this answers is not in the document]
remyb2canada: :)
good point: 4 is a repeat of 3, could even just be a measurement of 3. 
I'll take out 4.

Strategies/IITB_KPIs.md
remyb2canada: All the indicators here are meant to be used IF our strategies are put in place.

This one is to push towards a service approach in that the complexity and growth of a service is of no concern to its consumer. Currently, our "service" is through BRDs and personnel planning. The complexity of our "service" is a concern to our consumer because they have to plan according to our demand. 

This is the fear that some have. E.g.putting an API on the API store "How will we manage growth? What if we open the flood gate and we can't keep up? What if what if what if ??" Service uptime is of the responsibility of the service owner and they'll need to handle that complexity. Cloud is great for its rapid elasticity, DevOps is great for rapid response to change, etc.

Strategies/IITB_KPIs.md
remyb2canada: Good point re Product Management.
Currently, IITB cannot meet demands. If you have a "project" you'll want to engage IITB with, they'll probably say "we have time for you in 2024". Our strategies will be to improve this demand fulfillment. Through initiatives like (for example) moving to Product Management, DevOps, Bi-Modal Governance, API-first design 

Would there be another metric that we can use to measure IITB's ability to improve its capacity? Even with product mgt and CD, there must be something tracked that we can use. Thoughts?

Strategies/IITB_KPIs.md
remyb2canada: +1 on the resilience
would it be 2 indicators then? One for reduction in incidents, one for improved resilience when incidents occur?

Strategies/IITB_KPIs.md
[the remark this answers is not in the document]
gcharest: Might need to add references everywhere in our docs then but I'm good with this!

Strategies/IITB_KPIs.md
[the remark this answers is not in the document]
gcharest: Sounds good!

Strategies/IITB_KPIs.md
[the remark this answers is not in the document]
gcharest: OK, will need to review once again after we merge.

Strategies/IITB_KPIs.md
gcharest: Sounds good!

Strategies/IITB_KPIs.md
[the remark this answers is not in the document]
gcharest: OK, will craete a comment at bottom so we can merge

Strategies/IITB_KPIs.md
gcharest: Gartner must have something :wink:

Adding in a separate issue to discuss.

Strategies/IITB_KPIs.md
gcharest: Yes, that would make sense.

Adding in issue to track but merge this document in.

PR #36, opened 2019-07-03
New Strategy that asks to invest in 7 initiatives

remyb2canada:
Suggest these 7 initiatives be the priorities for DS:

1. Create an Enterprise Information Model
2. Accelerate Cloud Adoption
3. Invest in an Integration Layer (API-first approach)
4. Implement DevOps
5. Shift to Product Management
6. Adopt a Bi-Modal Governance and Delivery
7. Truly work in the open

I suggest we take 2 weeks to review because Gabriel and Seb are on vacation. 

Would also be worthwhile to have the SAs review it but I'd recommend a 2nd review (after our team has a chance to clean this up)


Strategies/TransitionToServices.md
```diff
@@ -0,0 +1,136 @@
...
+
+This goal will be achieved through the establishment and execution of the following **7 key strategic initiatives**:
+
+1. Create an Enterprise Information Model
+2. Accelerate Cloud Adoption
+3. Invest in an Integration Layer (API-first approach)
+4. Implement DevOps
+5. Shift to Product Management
+6. Adopt a Bi-Modal Governance and Delivery
+7. Truly work in the open
```
jaysonmc: Truly work in the open -> Increase Volume of those working in the open

#actionoriented ;)

Strategies/TransitionToServices.md
```diff
@@ -0,0 +1,136 @@
...
+
+Any privacy concerns will be treated separately, outside of the Enterprise Information Model.
+
+The Enterprise Information Model is part of a broader undertaking by IITB's Information Architecture directorate that seeks to mature ESDC's overall Information Architecture capability.
+
+## 2. Accelerate Cloud Adoption
+
+IITB will aggressively seek opportunities to adopt cloud computing and reap the benefits of rapid elasticity and increased agility.
+
+Priority will be given to SaaS and PaaS service models in a Public Cloud provider.
```
jaysonmc: Not sure what this is intended to mean ... priority over what? 

Could we summarive this in one sentence. Cloud first. 

?

Strategies/TransitionToServices.md
```diff
@@ -0,0 +1,136 @@
...
+
+Accelerating cloud adoption at ESDC will require the following:
+
+1. Any new software development must be developed *AND* deployed to the public cloud. This includes up to Protected B data classification.
+2. Development teams (dev, ops, testing) must be allocated a public cloud sandbox environment, one per team, with the following characteristics:
```
jaysonmc: not sure what "one per team means" ... 
How about ...

Each team must have a public cloud sandbox made available to them with no restrictions or connectivity to existing infrastructure, which is intended solely for learning, innovation, and experimentation

Strategies/TransitionToServices.md
gcharest: Updated to: Transition to Working in the Open by Default

Strategies/TransitionToServices.md
gcharest: I would actually move towards a Cloud Native organisation. That's what the industry calls it and it makes way more sense:

- Leverage Cloud services first
- If building on premise, you can build in a way that natively supports Cloud even though you may not already be able to do so yet.

Strategies/TransitionToServices.md
gcharest: Changed sentence to the following:
>Development teams (dev, ops, testing) must have access to a public cloud sandbox environment, with the following characteristics [...]

Strategies/TransitionToServices.md
gcharest: Removed the line in latest version of this document.

gcharest:
So, few things:
- 1 & 3 Are somewhat related; plus 1 is supposedly taken already by CDO. Could still keep and support though.
- 3 is very technical and I'm not sure they will "see" the picture
- 6 shouldn't be tied to dollar figure like the current projects onboarding. Should also encourage micro-services architectures, self-contained applications as much as possible.

remyb2canada:
Damned, I was making changes. Shouldn't have been merged yet

PR #42, opened 2019-07-05
Remove files

gcharest:
Cleaned up the repo of superflus files (they all can be found back in the history if needed)


gcharest:
@remyb2canada Can you please double check the files to be deleted and merge?

Thanks! :grin:

PR #54, opened 2019-07-10
adding strategy summary doc

jaysonmc:
Please review before approving. There is lots of missing information, I'm sure everything can think of at least one box or link to add. 

I think this is the kind of summarized documentation our senior management is looking for.


arushisharma19:
#ExperimentingwithGithub #DevelopingUnderstandingofITneeds #ThrowingBackQuestionsfromWhatILearntfromtheTeam  
Q1. Are the number on the left stages of development/stack?  
Q2. Where does APIs fit in the picture? Delivery of services? Is delivery of services the existing state (stack 1)?  
Q3. Looks like we are modernizing a lot of internal processes and development practices, not the solutions itself. Though, we are facilitating innovation through it. Isn`t HDLC about improving digital experience for the Canadians – no block talks about the solution or collaborating with others on solution?  
Q4. Tooling is mentioned over modern workplace and delivery of services – both?

gcharest:
@arushisharma19 Those are good questions, especially around the modernizing internal practices and tools vs the end-user solutions (Q3).

Part of this I think is that this would be a roadmap to that end? We can't currently change the solutions themselves since we don't have buy-in from business because we're not meeting demand on so many levels.

By addressing the first stage, we start to enable ourselves to tackle some of the larger problems?

Regarding Q4, maybe we need a better series of buckets, more closely tied in to our strategies (which we probably need to realign as well)

gcharest:
@arushisharma19 @jaysonmc 

Regarding Q1, maybe moving this picture from bottom -> up to left -> right if we're kind of talking about roadmaps?

Otherwise, it can be described as steps to the end state, I guess, and we should just remove the numbers to avoid this question by others lol

remyb2canada:
1. I don't know how to open a draw.io file from a Github branch :anguished: 
2. This is probably what Armand was looking for in terms of a picture of IT Strategy
3. Not sure what the numbers represent. Are they maturity levels? Going from 1. basic level, to 6 which is mastery?
4. What do colour codes mean? A legend would be ideal.
5. Need to add the following topics:
- talent management (HR)
- cyber security
- infrastructure (network, cloud, the idea of a production network != corporate network != internet)
- technical debt management
- data management (close relationship with CDO)
- open source (publishing and contributing to)
- API layer
6. For 3. Standardize tools (GCCode), this is going antipattern of allowing product teams to decide their technology stacks. I.e. Shaun's slack message about how managers wants to standardize dev tools already without having a clear idea of what is out there. Perhaps we mean standardize code repositories (plural, GCCode + GitHub/GitLab/Bitbucket)?

gcharest:
@remyb2canada to open it up, you need to do a checkout of the branch and update it:

```bash
git fetch -a
git checkout AddStrategySummary
git pull
```

Then you navigate to the folder and open up in draw.io

gcharest:
Agreed with @remyb2canada 

Should we map this to the TOGAF model? Isn't it the framework used here anyway? BIATSPA
Business (Services)
Information
Application
Technology
Security
Privacy
Accessibility

gcharest:
@jaysonmc propose we go forward with merge at this point?

jaysonmc:
@gcharest Works for me. I would like to make updates in a new PR, so would appreciate if we could merge this version for now.

PR #56, opened 2019-07-11
Changing to Agile Project Management > Product Management

jaysonmc:
Project Management won't be replaced by product management, though product management will increase in importance. 

Instead put an emphasis on shifting towards an agile project management framework


jaysonmc:
Wops. Let me run tests first. My mistake.

remyb2canada:
@jaysonmc  I disagree. Product Management is necessary and is different than Project Management (or Agile Project Management). Project Management will still exists but the shift towards Product Management has different implications than project management. Such as stable funding, continuous releases, dedicated teams, direct feedback from end users.

PR #63, opened 2019-07-16
Pages

gcharest:
Fixed the baseurl and update the gemfile


gcharest:
I don't think so but I'd like the site to be as portable as possible. Need to do some additional tests I guess.

Issue #70, opened 2019-07-17
Setup bilingual Jekyll site and Pages


smellems:
Use existing Pages as examples
- https://github.com/canada-ca/ore-ero
- https://github.com/smellems/smellems.github.io

Questions:
- How do we organize documents to be bilingual when posted.
- The current theme defines the default layout, we would need to change it to have bilingual title and description as well as to have control of the header and footer.  Do we want to use WET?

PR #72, opened 2019-07-17
Add French home page and toggle to switch language

smellems:
Start to #70


smellems:
Spell check doesn't like French..

gcharest:
Need to configure properly and might want to setup folders accordingly.

smellems:
Also added the favicon link

smellems:
#71

PR #73, opened 2019-07-18
Mandate

gcharest:
Had a quick meeting yesterday with @remyb2canada @jaysonmc and @smellems to try to clarify the mandate.

Please review to make sure it's properly set based on feedback.

Thanks!

### DIFF
--- a/TeamMandate/Mandate.md
+++ b/TeamMandate/Mandate.md
@@ -8,13 +8,10 @@ The [Canadian Experience](../Strategies/HumanDevelopmentLifeCycle.md) is when Ca
 
 Be the catalysts of change within IITB to lay the path for the evolution of ESDC's digital services.
 
-This will be done through an open collaborative approach that will further ESDC's mandate in the digital era by aligning IITB with the Government of Canada's Digital Standards and Treasury Board's Policy on Service and Digital.
-
 ## Mandate (team's authority within IITB)
 
-- Define the vision of IT within ESDC
-- Briefing senior management on key organisational problems and recommended steps to address them
-- Present recommended course of action to any ESDC governance body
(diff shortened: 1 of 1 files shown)


TeamMandate/Mandate.md
```diff
@@ -8,22 +8,15 @@ The [Canadian Experience](../Strategies/HumanDevelopmentLifeCycle.md) is when Ca
...
-This will be done through an open collaborative approach that will further ESDC's mandate in the digital era by aligning IITB with the Government of Canada's Digital Standards and Treasury Board's Policy on Service and Digital.
-
 ## Mandate (team's authority within IITB)
 
-- Define the vision of IT within ESDC
-- Briefing senior management on key organisational problems and recommended steps to address them
-- Present recommended course of action to any ESDC governance body
+- Diagnose key ESDC organisational problems
+- Identify recommended course of action to any ESDC governance body
+- Measure progress of strategies implementation against approved key performance
```
jaysonmc: I would add "ability to review information from groups within ESDC as they pertain to efficiency" or some such thing. 

ie. If we ask for information on how long someones process takes, and they don't give us access, we have something to point to that says we should in fact be granted access (granted it should be in the open anyways, but that's not the reality we work in ATM).

gcharest: Ah OK, I see! Trying to come up with a sentence simliar to the others.

gcharest:
@gabrielcossette as-tu eu une chance de réviser?

TeamMandate/Mandate.md
gabrielcossette: In my view, "Diagnose key ESDC organisational problems" gives us this authority.

PR #141, opened 2019-07-25
Mandate - Convert to presentation and merge with old presentation

smellems:
I converted the mandate page to a presentation.  I kept the content from the old presentation that we used last time.

Please review and comment before we add the French version


smellems:
Main file to review: https://github.com/sara-sabr/ITStrategy/pull/141/files?file-filters%5B%5D=.md&hide-deleted-files=true#diff-e7502f8b54b38fc9433ed69d88900700

If your running Jekyll locally you can view the presentation here
http://localhost:4000/ITStrategy/presentation.html?markdown=en/2019-07-xx-Mandate.md

presentations/en/2019-07-xx-Mandate.md
```diff
@@ -1,47 +1,128 @@
...
 ## Vision
 
-### Future Testimonials
+### Image of the future
+
+The [Canadian Experience](https://github.com/sara-sabr/ITStrategy/blob/master/Strategies/HumanDevelopmentLifeCycle.md) is when Canadians are benefiting seamlessly of the federal government services through modern technologies.
+
+--
+
+### Testimonials
```
gcharest: I don't understand this section and we probably need user research to confirm but my gut feeling tells me this is going to be common.

gabrielcossette:
Is it possible to make the slide separators "invisible" so that they don't show when we view the document in non-presentation mode?

It would be prettier as we'll have a single doc.

presentations/en/2019-07-xx-Mandate.md
```diff
@@ -1,47 +1,128 @@
...
 
 I just lost my job. It's been a month since my last paycheque and today I received a deposit of $1245.28 in my TD checking account. With this, I was able to buy groceries for my family while job hunting full time.
 
 As a student who isn't taking any courses this summer, I decided to look for work. My Government of Canada LinkedIn friend just sent me the perfect job opportunity to be an intern at Shopify. I'm excited for my interview next week!
 
 ---
 
+## Mission
```
remyb2canada: change to Mission (reason for existence)
and remove sub-section "Team's reason for existence"

presentations/en/2019-07-xx-Mandate.md
```diff
@@ -50,41 +131,51 @@ Federal public servants are entrusted to use and care for public resources respo
...
 ### Avoid Toil
 
 Toil is the kind of work tied to running a production service that tends to be manual, repetitive, automatable, tactical, devoid of enduring value, and that scales linearly as a service grows.
 
+--
+
 #### Repercussions
 
 - Re-define success
   - https://twitter.com/STP_PTS/status/1144268119025885186
```
remyb2canada: euhm... should we add this on a public forum?
perhaps keep this in our backpocket only.
Instead, success should be :
re-define success: when ESDC's lines of businesses succeed, IT succeeds.

presentations/en/2019-07-xx-Mandate.md
```diff
@@ -50,41 +131,51 @@ Federal public servants are entrusted to use and care for public resources respo
...
 
+--
+
 #### Opportunities
```
remyb2canada: add
- blockchain
- AI
- video robots

@jaysonmc  do you agree?
:laughing:

presentations/en/2019-07-xx-Mandate.md
smellems: I changed the `(...)` to a subtitle because it's since this is a presentation.  I think it's a good idea to remove mention of "team's", I removed it from other subtitles.

We could keep the `(...)` if we want.

presentations/en/2019-07-xx-Mandate.md
smellems: Or add the quote to the presentation. It was there, but we can take it out.  The whole last section needs more review.

presentations/en/2019-07-xx-Mandate.md
smellems: I put it with the vision, I don't know..  Take it out?

gabrielcossette:
If the goal is still to have a single doc that serves as doc + deck, it should also be generated as a webpage. It is not currently if I run jekyll serve. Just as an FYI.

gcharest:
We can't do both as it is but we can explore options.

I think we just need to not duplicate where our stuff is. Either or I
guess.

presentations/en/2019-07-xx-Mandate.md
gcharest: I think we may be combining two different things here? A presentation in my mind is a message to a specific audience.

Having our mandate page is a go-to place for a quick understanding of what we're doing.

jaysonmc:
I think it's awesome like this. Yeah, the mandate md file looks a little wonky, but who is reading that really? I mean, uh and maybe some other geeks. But by and large the only thing business or our management is concerned with is how it displays on the website, no? 

So putting up with a little wonkiness in the document is worth having a single source of truth, I suggest.

gcharest:
I'll double check the last changes and we can merge. I guess I'm just
worried we're trying to jam two things into one for the sake of single
source of truth. Sometimes we just need to take a few steps back to see
what doesn't really work. 😄

remyb2canada:
> 
> 
> I think it's awesome like this. Yeah, the mandate md file looks a little wonky, but who is reading that really? I mean, uh and maybe some other geeks. But by and large the only thing business or our management is concerned with is how it displays on the website, no?
> 
> So putting up with a little wonkiness in the document is worth having a single source of truth, I suggest.

presentation != web publishing though?

md file would be the web publishing
presentation is meant to be very concise and... well.. presented :)

gcharest:
I think this needs to be both a page (short version) and a presentation (telling the story)

smellems:
The initial idea that everything be a presentation is not proving to be very popular.

Issues raised
- Weird formatting when looking at markdown files on Github (slide seperators)
- Presentation does not allow for quick overview of documents (swipe, click, n)
- Presentation is not a good way to publish content online (similar to above)

Then we should put all documents in `_pages` and create presentations only when we will be presenting something somewhere.  pages will have all the details and presentations would be the short version often asking for something.

I think we can just close this PR.  We already have a mandate page and presentation in master.

We might also want to delete this repo, where I got the longer mandate presentation content: https://github.com/sara-sabr/presentations

gabrielcossette:
I share your view that we should keep pages and presentations separate for now.

Some more thoughts:
1. Are the changes to the Mandate presentation still worthwhile? If so, we could adjust and still merge this PR.
1. Also in favour of deleting presentations repo but we should check if this other version of Mandate presentation is useful to keep (or replace our other version(s)).
1. Nonetheless, we should keep the addition of the automatically generated list of presentations in this PR.

smellems:
The rebased branch now includes the updated presentation (with the content from the other repo) and the current mandate page.

PR #145, opened 2019-07-27
Rename CONTRIBUTING.MD to CONTRIBUTING.md

gcharest:
Rename extension file (was put in caps but should be lower case.)


gcharest:
Most links pointed to md, not MD which was breaking links

PR #146, opened 2019-07-29
Enable learning

jaysonmc:
Still work required (hence in wip-tec) 

First attempt to starting the document for our enable learning block in the strategic roadmap


gcharest:
An additional point is with regards to the DevOps Handbook of creating safe place for learning:

Blameless review of our incidents lead to an organization where we can continuously improve.

In a way, learning can be through courses, books and online resources but also through the investigation of issues/incidents. The lessons learned need to be shared throughout the organization but these are nowhere to be found at the moment, all tucked away.

jaysonmc:
@gcharest This is a good point. Think we need to add the redefinition of working. Even taking time to learn who is doing what in your org (or what your clients are doing), could do wonders. 

I'll add in something addressing this

gcharest:
@jaysonmc just need to fix the conflict otherwise I'd approve the submission

PR #156, opened 2019-08-01
Pages

gcharest:
Added the KPI page and fixed dead links


gcharest:
We need to rebase because changes were made to KPI file before deletion on this branch. Will fix today.

PR #166, opened 2019-08-08
Preabme for our planning documents

jaysonmc:
- Creating a layout that can be subtituted into our medium and long term plans 

- More information and data needed


smellems:
Should this be in `_pages`?  or will it be included in multiple places?

jaysonmc:
@smellems I was thinking we could include it on the page for both medium term and long term, instead of copity pasta or writing two separate ones

PR #184, opened 2019-08-14
Fix TOC + anchor scroll

gabrielcossette:
- Fix table of content links (introduced by a bug in my VS Code plugin)
- Add JS so that when we click on an anchor (e.g. in table of content), the scroll is not hidden by the fixed header


smellems:
Do we need (want to add) all this Javascript just to not have the page hide behind the menu?

I think this could be fixed by changing the HTML for the site.  Something with the navbar, maybe
https://getbootstrap.com/docs/4.0/components/navbar/

gabrielcossette:
Good point. I looked at more solutions and found one in pure CSS. We make the header non scrollable. PR has been adjusted.

_pages/en/it-picture-long-term.md
```diff
@@ -13,16 +13,17 @@ This is a working document to help create a picture of what will IITB (Innovatio
...
-  - [Automatic Enrollment and Delivery](#Automatic-Enrollment-and-Delivery)
-  - [Self-Service](#Self-Service)
-  - [Service Centers and Outreach](#Service-Centers-and-Outreach)
-- [Private Sector](#Private-Sector)
-  - [Any platform, any device, 3rd party integration](#any-platform-any-device-3rd-party-integration)
-  - [Open Collaboration](#open-collaboration)
-- [IITB-ESDC](#IITB-ESDC)
-  - [Modern Open Organization](#Modern-Open-Organization)
-  - [Strong Culture of Learning](#Strong-culture-of-learning)
+- [IT Vision Picture - Long Term](#it-vision-picture---long-term)
```
smellems: Do we need an anchor to the page title?

PR #194, opened 2019-08-20
Long term IITB vision update

smellems:
Moved things around and added headers under modern open org based on capabilities.  Also removed numbers %

### DIFF
--- a/_pages/en/it-picture-long-term.md
+++ b/_pages/en/it-picture-long-term.md
@@ -49,9 +49,12 @@ Results in easier access to services through delivering services to the client t
 
 #### Service Centres and Outreach
 
-For the 10% not enrolled automatically and unable to use the GC Portal (3 700 000 people), small service centres are available, especially in remote regions.
-Their role is to help citizens register and use digital services.
-They can also use the digital services on behalf of clients.
+For the 10% not enrolled automatically and unable to use the GC Portal (3 700 000 people), small service centres are available, especially in remote regions. Further, the costs saved through reduced real estate could be invested in sending representatives to areas where service uptake is lagging. In this way, through a reduced reliance on physical real estate, ESDC is better able to service regions in need. 
+
+Said service representatives would be able to... 
+
+- Help citizens register and use digital services
+- Use the digital services on behalf of clients (there should be no separate 'admin service functions', these should all be automated)
 
 GC departments have common kiosks to offer services, such as passports, taxes, grants as well as benefits from ESDC.
 Outreach programs are deployed to raise awareness in low uptake areas.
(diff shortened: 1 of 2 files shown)


_pages/en/it-picture-long-term.md
```diff
@@ -87,35 +87,59 @@ Ecosystem of services are created by leveraging private sector expertise where o
...
 
-IITB is a modern open organization with a very open culture.
+##### Enterprise Architecture
+
+Enterprise Architecture (EA) committees and review boards at IITB-ESDC have aligned with TBS mandatory procedures for EA (users, open standards, open source software and cloud first).
 Wherever possible code and projects are developed in the open and use of open source software is the norm.
-IITB is the digital, IM and IT advisor and enabler for ESDC and part of major business decisions.
-IITB is responsible for providing the automated services required to run a secure and agile business.
-This includes, automated software onboarding approval, key data exchange systems with which business may interact with, and an environment by which business may securely deploy their own solutions.
+EA teams enforce a loosely coupled architecture that empowers development and operations teams to meet business needs.
```
jaysonmc: Instead of  'enforce' I would say provide. 

Example: EA is responsible for managing our artifact repository. If you use all libraries from said repository then you are compliant. If not, you are not. 

Each group has the option to use or not use the libraries. EA is measured against the number of people who opt into compliance.

_pages/en/it-picture-long-term.md
```diff
@@ -87,35 +87,59 @@ Ecosystem of services are created by leveraging private sector expertise where o
...
-  - As a stewart, IITB provides ESDC Programs and their employees with the following services:
-    - Corporate services (hardware and software support)
-    - Cyber security
-    - Technology architecture
-    - Development teams
-- Automation of internal services
-- Leveraging IT services (from IITB) in a self-serving method
+Internal and external services are developed and supported as APIs first.
+APIs for services to Canadians are made available through the API store.
+All user interfaces (UI) developed use existing APIs and don't interact with the database directly.
```
jaysonmc: "and don't interact with the database directly." ... Goodness, I certainly hope so. Feel like we should be able to remove this?

remyb2canada:
Oh uh, I've also been making changes to long term view...

_pages/en/it-picture-long-term.md
remyb2canada: change 
All user interfaces (UI) developed use existing APIs and don't interact with the database directly.
to
User Interfaces (UI) are completely decoupled from the service interfaces. Allowing for different UIs to be built and interact with the service according to the channel  of choice of the end-user.

smellems: I added an "all" :man_facepalming:

Now removed

PR #201, opened 2019-08-22
Changing phrasing 

jaysonmc:
The way it was written implies a static perception of privacy and was too blanket. 

If we have a digital ID, we have a digital ID. And if you want to be paperless, we use that digital ID. So, for taxes or medical appts, that's just what you use, perhaps opting out is not a feature. 

For automated EI enrollement where the bank would exchange certain information with the government to be able to send you money, that could be an opt-out system. 

I would also propose it'd be best to make it an opt-out system rather than an opt-in one (if legally permitted, obviously. I'm talking from a digital effeciency perspective). My intuition would be, more people would want it, but not opt-in, than the amount of people who don't want it and would opt-out (the few technical people who care enough to do so). Of course, we should look to validate or invalidate said intuition.

### DIFF
--- a/_pages/en/it-picture-long-term.md
+++ b/_pages/en/it-picture-long-term.md
@@ -56,15 +56,13 @@ Results in easier access to services through delivering services to the client t
 
 #### Service Centres and Outreach
 
-For the 10% not enrolled automatically and unable to use the GC Portal (3 700 000 people), small service centres are available, especially in remote regions. Further, the costs saved through reduced real estate could be invested in sending representatives to areas where service uptake is lagging. In this way, through a reduced reliance on physical real estate, ESDC is better able to service regions in need.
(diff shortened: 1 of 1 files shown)


jaysonmc:
Would also suggest we put in language about internet accessibility throughout Canada. 

Increasing, but how quickly? Can we assume everyone can access the internet through technology like https://loon.com/ ?

_pages/en/it-picture-long-term.md
```diff
@@ -27,18 +27,24 @@ In the following sections, we're trying to take the same end state and create "v
...
 
 #### Automatic Enrollment and Delivery
 
-Enrollment and delivery of services is automated by default and wherever possible.
-These Canadians are benefiting seamlessly from GC services.
-GC Services are integrated with widely used third party platforms and services (e.g. Banks, CRA, Social media, Job boards, Internet of things).
-Services are accessible anywhere, anytime, from any device, and from any partner.
+Enrollment and delivery of services is automated by default and wherever possible, with an opt-out option.
+
+Canadians who do not opt out benefit seamlessly from GC services through the set of services they opt in to. For example their ...
```
gabrielcossette: Maybe rephrase "they opt in to" as it's an opt out system ;)

Issue #196, opened 2019-08-20
How to send feedback page


remyb2canada:
Add a reference (separate page or inline paragraph) to both IT Picture pages (medium and long term) describing how one can send feedback to IT Strategy team.

gcharest:
Related to #183 I think.

gcharest:
I think we can add this as part of the disclaimer? This would be easily done and would be applied to all pages across the site?

gcharest:
Need to be more generic title.

gcharest:
Similar to #183

remyb2canada:
add french version

Issue #183, opened 2019-08-14
Website Feedback 


jaysonmc:
- [x] Add Contact header 

1. Create Issue 
2. Email 
3. Submit Pull Request (for techies) 

- [x] Distribution mailbox

gcharest:
What's the contract header?

gabrielcossette:
@gcharest Typo. It's rather "Contact" header ;)

gcharest:
Ah, makes more sense now lol

Le jeu. 15 août 2019 11 h 04, Gabriel Cossette <notifications@github.com> a
écrit :

PR #242, opened 2019-09-09
Add Feedback notice

gabrielcossette:
An attempt at adding a Feedback notice header. Let me know what you think :)

Related to #183 and #196.


jaysonmc:
If we add that banner, should we remove it from the text as well? 

I like the idea of making it easier to quickily navigate. I just created an issue (#243) for creating a header, which I think is trying to address the same problem as this navbar.

remyb2canada:
I like a head on every page but that instead links to a "more details" page.
E.g. a "What is this about and how can I contribute to it?"

This header would be a hyperlink to a separate page explaining:
- about us / mandate
- how to contribute

similarly to the speaking_notes.md one but done more generaly
https://github.com/sara-sabr/ITStrategy/blob/master/_pages/en/speaking_notes.md

gcharest:
The notice works as proposed and I would approve it as is but I tend to agree with @remyb2canada proposal to link to the How To instead of just our repo.

PR #244, opened 2019-09-10
PLEASE TEST JEKYLL!

remyb2canada:
I'm trying to test the jekyll generated site using labs.play-with-docker.com but keep getting connection errors.

I've changed the permalink reference because of OCD.

- updates on what it means to work in the open and how to contribute
- generalized content to be usable for any documents (with the idea that
documents would contain headers linkable to this page)
- renamed file because of OCD


gcharest:
Having the web socket open would really help there 🤔

remyb2canada:
@gcharest, using Edge works. But I think my docker-fu isn't strong enough. 

On labs.play-with-docker.com, the resulting localhost Jekyll is translated to be accessible via internet. Adding a other level of complexity.

gcharest:
Update with a workaround page to link to new URL needed.

gcharest:
Last push included the following changes:
- Updated the French text for transition page
- Set status of both pages no longer displayed as `archived` to not render in "live pages"

PR #259, opened 2019-09-17
Categories

gcharest:
Updated the include listPages.md to automatically group by categories.
If no category is specified (needs to be a listed category), the page will not be displayed.


jaysonmc:
Do we need the top section on top of the table of contents? I would suggest we remove that. 

That was just brainstorming ... 

If we remove that I'd say it's good to go

gcharest:
@smellems Should we remove the pages status? I think that if a page is categorized, it behaves the same so it becomes a bit redundant, no?

smellems:
I think I made it simpler and removed the check for the status.  We could remove it from pages.

One thing I noticed is that we are using an `h4` inside the list.  We should probably change that since they are not headers.

PR #274, opened 2019-09-27
readded mention of APM

remyb2canada:
because it was removed in a previous commit. APM is expected to be used as a tool to manage technical debt (from our conversation with technical debt ppl)


remyb2canada:
Someone change the word APM for something else. While keeping wordings around mapping of apps to business services :)

Am off !!!

PR #280, opened 2019-10-02
Software onboarding wireframe demo

jaysonmc:
Initial wireframe for software onboarding demo


smellems:
I think we should mention NPM packages, since Node is already approved and they are like plugins.

- CLI Software to onboard
  - NPM
    - Enter name and version of package from npmjs.com
      - Security (npm audit)
      - OSS licence (OSI or FSF)
  - Other

We could also automate checks for Ruby and Python packages if those languages are approved

jaysonmc:
@smellems 
Okay, I specified that it's only npm for now. Can add others as we go

smellems:
For production or experimentation..  I think we should ask if the package will be used
- as part of a production application or to build the production application;
- for testing and experimentation (including testing of production applications)

Also, since this is about OSS do we want to ask if they will be modifying the code or assume that they wont?

gcharest:
So, this is pretty good and I was wondering about the reference to NPM only. What about everything else? 

Also, that's for the "initial" onboarding. Do we look into the tracking and maintaining of these dependencies?

gcharest:
And it's developer intended, correct?

PR #281, opened 2019-10-02
Updating references

jaysonmc:
- Adding podcasts 
- Adding learning resources


jaysonmc:
Wops ... one second :(

gcharest:
@smellems @jaysonmc  Could it have been run on the work laptop which doesn't have same version of Node.js?

Issue #295, opened 2019-10-09
IT Strategy Website - Accessibility Assessment


gcharest:

Moro-Code:
Assigning myself and Elmina to work on this

PR #307, opened 2019-10-15
Accessibility link color fixes

smellems:
Related to #295 I started testing pages for accessibility on the Website using [Pa11y](https://www.npmjs.com/package/pa11y).

Using docker you can test (to test this PR you need to host somewhere, like docker playground)

`docker run --rm frvge/pa11y https://sara-sabr.github.io/ITStrategy/home.html`

Pa11y only tests one page at a time, but this PR fixes errors that are on all pages.  The color of links in the notices and the footer
> Error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1.
>  ├── WCAG2AA.Principle1.Guideline1_4.1_4_3.G18.Fail

More complex pages still have errors to look at, ex.:
`docker run --rm frvge/pa11y https://sara-sabr.github.io/ITStrategy/strategy-summary.html`


gabrielcossette:
Great find! I'm all for Accessibility but we almost can't notice that there's a link with this CSS class. Can we increase the contrast ratio and keep it looking like a link?

smellems:
Updated link color in alert to use the style of the alert
https://getbootstrap.com/docs/4.0/components/alerts/#link-color

For the footer, since it's already a custom gray in the CSS, I suggest we change the color to use the `text-dark`.  The `text-primary` color is the same as the default blue

Issue #254, opened 2019-09-13
ITPSM - Enabling Team Strategy


jaysonmc:
Write a justification for our mgmt to submit to senior management to have all boxes under the director report directly to the director 

- More human way of organizing people
- More flexibility to assigning people to groups as needed (both for the employee and the org) 
- Need to address how PAs will work (done by manager, signed by director?) 
- Reference recommendations from IT Revolution books (Ex: Project to Product suggestions on Human Resources, and read Team topologies)

jaysonmc:
https://www.gartner.com/document/3957241?ref=QVREmail&utm_source=GartnerOther&utm_medium=email&utm_campaign=QVR&utm_content=QVR 
yeah that the guy i asked forcloud sandbox

jaysonmc:
Meeting upset for tomorrow afternoon to discuss (confirmation pending)

jaysonmc:
Updated name from: Justify Flattening Org to ITPSM - Enabling Team Strategy

PR #309, opened 2019-10-15
Request for team review: Enabling directorate

jaysonmc:
Proposal for how best to organize an enabling directorate

Work related to issue #254, requested by director

### DIFF
--- a/_pages/en/building-enabling-directorate.md
+++ b/_pages/en/building-enabling-directorate.md
@@ -19,13 +19,53 @@ In [Team Topologies](https://itrevolution.com/book/team-topologies/) they write 
 - Enabling Teams
 - Complex Subsystem Teams
 
-In our particular case, to use it as an example, the IT Strategy team is an enabling team. We are under an director who himself considers his role to be an enabling team.  Therefore, the IT Strategy team is an enabling team, within an enabling team, and has other sibling teams that are also enabling teams. That being the case, how would one synchronize activities of the entire directorate? Say the director were to offer their services to a team elsewhere in the organization. How best to manage and coordinate the work between the series of sub-enabling teams? That is the question this document is intending to answer.
+In our particular case, to use it as an example, the IT Strategy team is an enabling team. We are under a director who considers the directorate to be an enabling team.  Therefore the IT Strategy team is an enabling team within an enabling team, and has other sibling teams that are also enabling teams. That being the case, how would one synchronize activities of the entire directorate? Say the director were to offer their services to a team elsewhere in the organization. How best to manage and coordinate the work between the series of sub-enabling teams? That is the question this document is intending to answer.
 
 ### Proposal
 
 #### Organizational
 
+##### Internal Interactions
+
+The teams may remain in their present functional structure for now, which aligns well with the governments expected structure for the sake of processes such as the Public Service Performance Management (PSPM). In our particular case, the teams will be as follows
+
+- IT Strategy
+- IT Service Management
+- Software Vendor Management
+- Hardware Vendor Management
+
+Each team has it`s own set of responsibilities for which they are responsible. The teams will collaborate regularly both via regular meetings and open collaboration tools (where possible and appropriate). For example, in addition to e-mail, tools such as Slack or Skype messenger should be used for regular informal communications and kanban boards should be open by default enabling each team to follow the progress of one another.
+
+##### External Interactions
+
+This is an initial proposal and is intended to allow for iterations to refine and modify the process as needed.
+
+###### Interactions
+
+The book Team Topologies, referenced below, recommends that Enabling Teams work with other teams for a period (potentially a few months) [TO-DO: Quote needed] in order to assist them with onboarding new technologies and approaches, and improving their existing processes and procedures. The goal is to ensure the interactions are brief enough that the teams being enabled do not become dependent on the enabling teams. The process improvements should stand on their own, and should not involve [toil](https://landing.google.com/sre/sre-book/chapters/eliminating-toil/). That is to say, the new processes and procedures should aim to automate or improve the existing processes, not lend temporary person-power to the teams. Given the short and iterative nature of the interactions between the teams (durations of interactions should be defined by number of sprints) the enabling teams must identify quick wins -- the amount of automation that could be done within the limited timeframe that would result in the greatest amount of automated work.
+
+###### Tooling
+
+[TO-DO: Quote from Team Topologies regarding varying levels of team communication]
+
+The amount of collaboration between the enabl*ing* teams and the enabl*ed* teams should vary significant during the time which the enabling team is paired with the enabled team. As such, collaboration tools that allow for the rapid creation and disintegration should be leveraged. An example of a viable tool for such a requirement would be Slack, where groups can be created and disposed of with relative ease.
+
 #### Process
+
+##### Teams Iterative Process
+
+- All teams within the directorate have synchronize sprints
+- Representatives from each enabling team within the enabling directorate meet prior to the synchronized sprint planning for the director to set direcorate level priorities
+- During the synchonrized sprint planning meetings ensure directorate level prioties are broken down into manageable tasks
+
+##### Engaging the Enabling Teams
+
+- Potential teams to be enabled should contact the enabling teams
+- An item is created on the kanban board of the enabling team
+- Representative from an enabling team meets with potential client team to prioritize work
+- Work is planned and prioritized by the enabling team during sprint planning
+
+*Note: It is important that the enabling teams have open kanban boards so that teams who have requested their attention can see where their tasks have been prioritized relative to others, or can see work in progress as the interaction takes place between the two teams.*
 
 ### References
 
--- a/_pages/en/building-enabling-directorate.md
+++ b/_pages/en/building-enabling-directorate.md
@@ -36,6 +36,16 @@ The teams may remain in their present functional structure for now, which aligns
 
 Each team has it`s own set of responsibilities for which they are responsible. The teams will collaborate regularly both via regular meetings and open collaboration tools (where possible and appropriate). For example, in addition to e-mail, tools such as Slack or Skype messenger should be used for regular informal communications and kanban boards should be open by default enabling each team to follow the progress of one another.
 
+###### Time Management
+
+Teams should not be fully allocated to their enabling function.
+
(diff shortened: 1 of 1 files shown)


_pages/en/building-enabling-directorate.md
```diff
@@ -0,0 +1,82 @@
...
+### Problem Statement
+
+In [Team Topologies](https://itrevolution.com/book/team-topologies/) they write that organizations should organize themselves by nesting the four team definitions.
+
+- Value Stream Aligned Teams
+- Platform Teams
+- Enabling Teams
+- Complex Subsystem Teams
+
+In our particular case, to use it as an example, the IT Strategy team is an enabling team. We are under a director who considers the directorate to be an enabling team.  Therefore the IT Strategy team is an enabling team within an enabling team, and has other sibling teams that are also enabling teams. That being the case, how would one synchronize activities of the entire directorate? Say the director was to offer their services to a team elsewhere in the organization. How best to manage and coordinate the work between the series of sub-enabling teams? That is the question this document is intending to answer.
```
remyb2canada: Would be worthwhile to have a one-two liner on what is an enabling team. e.g.
"those which are more expertise-focused. They facilitate other teams to do their work and to gain capabilities and be able to do all the things [like] testing, monitoring, deployment, etc., without requiring experts in each of those areas"

remyb2canada: Clarify that the scope of this document is focused on one directorate: ITSM (or IT Service & Vendor Management).
Unless the scope is that of the IT Strategy team towards the rest of IITB?
Or is the scope that of the ITSM team towards the rest of IITB?

_pages/en/building-enabling-directorate.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Proposal
+
+#### Organizational
+
+##### Internal Interactions
+
+The teams may remain in their present functional structure for now, which aligns well with the governments expected structure for the sake of processes such as the Public Service Performance Management (PSPM). In our particular case, the teams will be as follows:
+
+- IT Strategy
```
remyb2canada: I think we should state that the IT Strategy team's enabling value is aligned to their mandate (e.g. de-risking the use of IT. For example, identifying waste and excessive work in progress within the overall IT value chain).

_pages/en/building-enabling-directorate.md
```diff
@@ -0,0 +1,82 @@
...
+
+[TO-DO: Quote from Team Topologies regarding varying levels of team communication]
+
+The amount of collaboration between the enabling teams and the enabled teams should vary significantly during the time which the enabling team is paired with the enabled team. As such, collaboration tools that allow for the rapid creation and disintegration should be leveraged. An example of a viable tool for such a requirement would be Slack, where groups can be created and disposed of with relative ease.
+
+#### Process
+
+##### Teams Iterative Process
+
+- All teams within the directorate have synchronized sprints
```
remyb2canada: That seems prescriptive. Why not let different teams use their own cadence?

_pages/en/building-enabling-directorate.md
```diff
@@ -0,0 +1,82 @@
...
+
+###### Tooling
+
+[TO-DO: Quote from Team Topologies regarding varying levels of team communication]
+
+The amount of collaboration between the enabling teams and the enabled teams should vary significantly during the time which the enabling team is paired with the enabled team. As such, collaboration tools that allow for the rapid creation and disintegration should be leveraged. An example of a viable tool for such a requirement would be Slack, where groups can be created and disposed of with relative ease.
+
+#### Process
```
remyb2canada: Am I reading this correctly that this is an intake and WIP management process?
I.e. each ITSM teams (IT strategy, s/w, h/w, and IT service) have part of their time associated to enabling other teams. That means they have limitted capacity to intake enablement requests.

During the teams iterative process, each team will have to determine if they have capacity to take on new requests, or if their existing requests need to be re-prioritize.

_pages/en/building-enabling-directorate.md
```diff
@@ -0,0 +1,82 @@
...
+
+##### Teams Iterative Process
```
remyb2canada: I would add another item:
- If an enabling team has a particular task that has remained on the KANBAN board for too long (e.g. 2 months), that task goes back in the backlog and may be picked up by another enabling team

_pages/en/building-enabling-directorate.md
```diff
@@ -0,0 +1,82 @@
...
+
+- All teams within the directorate have synchronized sprints
+- Representatives from each enabling team within the enabling directorate meet prior to the synchronized sprint planning for the director to set directorate-level priorities
+- During the synchronized sprint planning meetings, ensure directorate-level priorities are broken down into manageable tasks
+
+##### Engaging the Enabling Teams
+
+- Potential teams to be enabled should contact the enabling teams
+- An item is created on the kanban board of the enabling team
```
remyb2canada: Maybe worthwhile to have a single kanban board for the whole directorate. A single "enabling" kanban board so that WIP and backlog is visible to the director in one place and has an easy monitoring ability to identify work that has been "in progress" for too long.

This is similar to the lean model where an hypothesis is produced and the validation of that hypothesis is not expected to take months otherwise it may indicate the hypothesis is false.

gcharest: This may be hard to manage but we could attempt it. Roll up of the cards may be a better option (not sure if there are solutions for this already).

_pages/en/building-enabling-directorate.md
gcharest: Agreed, or a definitions section might be useful. We need to keep in mind most people reading our content will not have a similar background. Just linking books will not cut it 😀 (A lot of people don't have time/interest to read work related books)

gcharest: My understanding is that IT Strategy is an enabling team, same as each other teams in ITSM. So that makes our directorate an enabling directorate (a team of enabling teams).

_pages/en/building-enabling-directorate.md
gcharest: "And establishing strategies to address those issues"

_pages/en/building-enabling-directorate.md
gcharest: Hmm, unless it's planning at the directorate level? Not sure about this here.

_pages/en/building-enabling-directorate.md
gcharest: I think since we are enabling teams, we don't have as much "requests"? Like, we get the odd off the wall request to do X but usually our role is to help other teams like the Value Stream Aligned Teams, which deal with requests?

_pages/en/building-enabling-directorate.md
gcharest: That should be discussed automatically at the sprint planning, every two weeks. If it hasn't been completed, it should probably even be discussed at the retrospective.

_pages/en/building-enabling-directorate.md
gcharest: Only thing I would caution is about redundant work. Need to avoid creating multiple "boards" to track the same things twice or more.

_pages/en/building-enabling-directorate.md
jaysonmc: I tried to make it general so other teams could leverage it as well. Though @gcharest was right ..

Unless the scope is that of the IT Strategy team towards the rest of IITB? Yes
Or is the scope that of the ITSM team towards the rest of IITB? Yes

SARA is an 'Enabling Team'
ITPSM is an 'Enabling Team'
The teams within ITPSM are 'Enabling Teams'

That's the power of the model in the book is that the definitions get recursively applied. 

But the focus on the article is how ITPSM will interact. And in that model I also attempt to describe how the teams will interact with each other, and other teams, as part of that model.

_pages/en/building-enabling-directorate.md
jaysonmc: I seem to recall a discussion where we suggested if folks are just re-wording or clarifying they should just modify the PR themselves. 

Discussion needed when folks would like to change the meaning of the idea/concept :)

_pages/en/building-enabling-directorate.md
jaysonmc: Because for the sake of coordinating work between teams, it would be best if they were all working on the same coordinated tasks at the same time. 

If work comes down from the director to the managers, if ones team sprint planning just happened, and the work was intended to take only one sprint, that means if the sprints are malaligned they will not be able to accomodate the work without having a mid-sprint planning session.

TLDR; it is to enable easier coordination between the teams under the same director.

_pages/en/building-enabling-directorate.md
jaysonmc: Could be a value stream aligned team, could be another enabling team. Could be anyone who wants to work better.

The idea is requests come from two places, and the team must prioritize how they manage the work during their planning meetings. 

Work may come from the director, if the entire directorate is enabling another directorate, for example. 

Work may come from one manager to another manage within ITPSM. 

In either case, the team must prioritize the work accoringly. 

In the case where the work comes from the director, this is when the synchonrized sprints come into play, as it will be hard to unify and coordinate work between 3 teams if they all out of sync when they are working on the work handed down by the director.

_pages/en/building-enabling-directorate.md
jaysonmc: @remyb2canada 

"picked up by another enabling team" 

hmm .. that would be tricky, I imagine. Possible, but not sure I'd vote in favour of such a process (not commonly, any who). 

Perhaps we could discuss how that mechanic could work... have trouble imaging it atm

_pages/en/building-enabling-directorate.md
jaysonmc: This is the idea of the synchronized sprints. 

Director meets with representative of teams
Teams discuss who will do what
Teams all have their sprint planning
Teams prioritize the work that they have taken on as part of the meeting with the director
Entire directorate completes sprint
Repeat (what did we all get done, what remains to be done)

So perhaps some master board with large tasks you can link to? Or perhaps milestones, if we all have our own projects or some such thing? I see the use case attempting to be solved here ... just not sure what tools would fit this requirement ... it's a neat idea though. 'sub kanban boards' 

Granted, that is supposed to be the responsibility of the product owner. Ensure they understand the greater priorities and goals and ensure that they're not doing double work and so forth. So the nested kanban board approach sounds like another way of solving the same problem the meetings with the director is supposed to solve (coordinating work between the teams in the directorate)

Issue #322, opened 2019-10-17
OSS Course - Hands-on Exercises


gcharest:
Dependency to #306

gcharest:
This is the series of exercises required to support the course content built here:
[Using OSS](https://sara-sabr.github.io/ITStrategy/presentation.html?markdown=/en/2019-11-12-OSS-Using.md)
[Publishing OSS](https://sara-sabr.github.io/ITStrategy/presentation.html?markdown=/en/2019-11-12-OSS-Publishing.md)
[Contributing to OSS](https://sara-sabr.github.io/ITStrategy/presentation.html?markdown=/en/2019-11-12-OSS-Contributing.md)

PR #333, opened 2019-10-18
Exercises for Using OSS course

gcharest:
Modified the Using OSS hands-on exercise based on user feedback.

Looking for both feedback into how it runs before merging.


gcharest:
OK, so I've moved the pages to... pages! Added the front matter and it's now ready to publish. Next PR will be:

- Changing Categories to Sections
- Adding the Exercises section to the `_config.yml`

gcharest:
@smellems Do you think blog posts would be better than pages?

gcharest:
Fixes #322

PR #339, opened 2019-10-21
re-added risk in the mission statement

remyb2canada:
Our mission statement must have de-risking the use of IT
It's the whole "transformation" in the digital transformation. As mentioned in War & Peace & IT

quote from page xxvi

> Now is the time to start transforming, because it's both low risk and urgent. You might be surprised that I say it's low risk, but I insist. Enterprises may feel like they should move slowly and cautiously, stepping onto the digital path only after they've checked the traffic coming from all directions.

> But the very point of digital transformation is to *reduce* risk. 

> Digital enterprises set risk-mitigating guardrails in place, then use their speed as a way to quickly adjust course when new risks appear.

### DIFF
--- a/_pages/en/mandate.md
+++ b/_pages/en/mandate.md
@@ -18,7 +18,7 @@ Continuous improvement transformation and GC wide "Government as a service" base
 
 ## Mission (team's reason for existence)
 
-Enable the strategic value of IT within ESDC by reducing its risks to increase business agility.
+Enable the strategic value of IT within ESDC by reducing its risks and increase business agility.
 
 ## Mandate (team's authority within IITB)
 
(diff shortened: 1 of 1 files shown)


_pages/en/mandate.md
```diff
@@ -18,7 +18,7 @@ Continuous improvement transformation and GC wide "Government as a service" base
 
 ## Mission (team's reason for existence)
 
-Enable the strategic value of IT within ESDC to increase business agility.
+Enable the strategic value of IT within ESDC by reducing its risks to increase business agility.
```
smellems: Can we say?
> .. by increasing business agility to reduce risks.

PR #357, opened 2019-10-24
List presentations based on files

### LINKED ISSUE #345: List presentations dynamically using Jekyll
Current method of listing presentations on our working documents site (https://sara-sabr.github.io/ITStrategy/documents-and-presentations.html) is hard-coded.

Would need a similar framework like the pages to lists dynamically the presentations that are status = posted

smellems:
Fixes #345

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -32,9 +32,6 @@ index:
 language:
   en: English
   fr: Français
-mandate:
-  en: Mandate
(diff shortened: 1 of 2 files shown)


smellems:
It currently list only English on English page and French on French page

gcharest:
OK, let's change in the future if we need.

PR #359, opened 2019-10-25
Update GC IT direction section - move some things

smellems:
other than changing the first section, I reordered to put the general statements before the "IT teams must.." statements


gcharest:
Rerunning the test passes. @smellems you can merge ;)

PR #365, opened 2019-10-25
adding suggested changes to presentation

jaysonmc:
- Changed definition/example of toil
- Adding flexibility as to how teams dedicate their time to improvement and training (don't need to say 10% here and 10% here, let people self organize for what works for their team in their situation) 
- Removed weird stuff at the end of SLF.md
- Diagnosing lead time was mingled in with negative things, making lead time sound like a negative, so reworded that


presentations/en/2019-11-05-CIOEC-Mandate.md
```diff
@@ -8,8 +8,8 @@ For discussion
 
 ## Context
 
-- The new Policy on Service and Digital requires government services to be delivered using technology
-- Demands for technology will increase exponentially
+- The new Policy on Service and Digital calls for a digital-first and digitally enabled government
```
remyb2canada: @jaysonmc 
From https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32603

> Overall, the Policy on Service and Digital advances the delivery of services and the effectiveness of government operations through [...] leveraging of information technology.

I'd like to state technology as opposed to digital. Because for our mandate, digital is a buzzword IMHO. Our mandate should focus on IT.

jaysonmc: Hmmm ... I understand this, however I wouldn't say that the new policy 'requires' it. They even temper the statement in the Digital Operations Strategy 

https://www.canada.ca/en/government/system/digital-government/digital-operations-strategic-plan-2018-2022.html

"A digital-first and digitally enabled government that is available anytime, anywhere, through any service window. This does not mean “digital only,” but it does mean that digital service delivery cannot be an afterthought and importantly, in-person and telephone interactions must be digitally enabled to deliver excellence in service."

Given that this section is alluding directly to TBS (rather than ourselves), I wanted to be sure we quote their documentation as directly as possible

Maybe Im splitting hairs, but we could use 'technology' instead of 'digital' but I wouln't say 'requires' (given it explicitly says 'This does not mean “digital only”')

jaysonmc: @remyb2canada 
Changed to ... 
'- The new Policy on Service and Digital calls for a digitally enabled government which delivers services through technology'

remyb2canada: kewl

Issue #360, opened 2019-10-25
ESDC IT Strategy Overview


remyb2canada:
Write a document that tells the story of what the IT Strategy team is looking to accomplished from its mandate, IT Pictures, Strategies and Actions, as well as its Strategy Map within the context of the shift towards Government as a Platform.

The document is to provide readers, expecting a single strategy document, with an understanding of the team's focus and approach to strategy making.

PR#367

PR #367, opened 2019-10-25
Overview of ESDC IT Strategy

remyb2canada:
Relates to issue #360 
Words on a page...

WARNING: Jekyll test not done

Purpose is to serve as central starting point of what a "single ESDC IT Strategy" would be. Even though we don't have a single one. As a means to respond to AG's feedback that we need to be better able to communicate our approach.

The content is a collection of work we did and ties it with a bow. Would be under the "About Us" section of our site.


remyb2canada:
> How long or how detailed do we want to make this?

@jaysonmc  that will be the trick! Somehow we must get it through that it's not a new technology or functional map that is the problem, it's how the conventional method of doing IT needs to be reviewed.

gcharest:
Build won't work, need to approve #371 to fix.

Otherwise, really like the content. Not too long, not too short either, ties everything together.

remyb2canada:
**BAD COMMIT** 
46d8af2 

do not merge until fixed

remyb2canada:
> 
> 
> Undo last commit. Looks like a merge after the rebase..
> 
> `git reset --hard HEAD^`
> 
> Force push the branch
> 
> `git push -f origin ESDCITStrategy`

done!

PR #375, opened 2019-10-28
Exclude gccode URL from test

gcharest:
Fix test on gccode which is on Intranet only.


gcharest:
@smellems, @remyb2canada proposed we could whitelist some of the known SaaS solutions that seem to be failing the tests like Youtube, Trello (today) as exceptions. We can set a review of these at certain intervals. Thoughts?

PR #376, opened 2019-10-28
Working open

gcharest:
This presentation is less technical. The intended audience will be for devs but also for data scientists, designers, etc. As such, there's less details than the other sessions.


gcharest:
Conflict introduced because I had improperly synched my local repo (working on too many computers!)

PR #398, opened 2019-11-05
Add bootstrap class for table and blockquote

### LINKED ISSUE #346: Markdown Code blocks not properly displayed in Bootstrap CSS
Using code blocks is not properly rendered in Jekyll with base Bootstrap CSS:

```bash
` ` `bash (no space in between the ` )
mkdir mydirectory
` ` `
```

Should be rendered something like this:

```bash
mkdir mydirectory
```

But right now, it generates the following HTML tags around the content but doesn't change anything except the font:
```html
<div class="language-bash highlighter-rouge">
   <div class="highlight">
      <pre class="highlight">
         <code>mkdir mydirectory</code>
      </pre>
   </div>
</div>
```

Probably need to look at padding and background color.

smellems:
I was testing tables on the website and found this solution to add missing classes.

Fixes #346


gcharest:
Just tested but didn't seem to trigger a change in display for code blocks. Works for blockquote and tables (adds the class) but need some work on the CSS to have them properly display.

I also added highlight.js plug-in to have code blocks properly displayed. Set it to a11y-dark.

Issue #399, opened 2019-11-06
Improve CSS for blockquotes and tables look and feel


gcharest:
Related to #398: Need to improve the way tables and blockquotes are displayed in the site.

PR #406, opened 2019-11-11
Change display of blog post list

gcharest:
Made a few changes to the way we display blog post list.

- Add date
- Add excerpt

Modified previous blog posts:

- Added  tag to identify part of the blog post should be part of the excerpt
- Removed the "##Summary" header so it would not be included in the excerpt for Case for Continuous Improvement" blog post

Created new blog post to explain the changes and the modified date on previous blog posts.

Regarding #399:

I initially used the block quote for the excerpt but removed it after testing a few options. For the purpose of the list, I think it wasn't useful.

However, I did reuse Mozilla's default block quote CSS and it looks pretty good now IMHO:

image


gcharest:
Just adding the previous way block quotes were being displayed for comparison purposes:

image

PR #455, opened 2019-11-25
Create Strategy presentation for 20% time +

smellems:
https://sara-sabr.github.io/util-presentation/presentation.html?url=https://raw.githubusercontent.com/sara-sabr/ITStrategy/strat-learn-auto-improv/presentations/en/Strategy-Continuous-Learning-Automating-Improving.md


presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
```diff
@@ -0,0 +1,141 @@
...
+- Try new technologies
+- Attend events and conferences
+- Stay up to date on trends and technologies
+- Share knowledge (IITB showcase, Dev CoP, ..)
+
+> Make it a requirement in performance agreements of directors, managers and team leads
+
+---
+
+### IT teams must use 10% of their time to automate and improve existing processes
```
jaysonmc: IT teams must use 10% of their time to automate and improve existing processes
=> 
IT teams must use 10% of their time to automate or improve existing processes

Some teams may not have expertise to automate ... 

Which, then again, in this case they should spend more time learning ... so should be really be diving these two? 10 and 10?

Just 20% and let the teams self manage how they use it would be preferable, no?

presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
```diff
@@ -0,0 +1,141 @@
...
+- Related to learning and experimentation
+- Teams must show progress on automation and continuous improvements
+- Add automation tools to Application Catalogue
+- Create centralized repository of scripts on GCcode
+
+> Make it a requirement in performance agreements of directors, managers and team leads
+
+---
+
+### IT teams must work using Agile methodologies
```
jaysonmc: I'd remove any buzz words which time bounds the strategies. Perhaps something like ... 

"Teams must work relentlessly to reduce lead times and increase responsiveness to business and operational needs"

presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
```diff
@@ -0,0 +1,141 @@
...
+
+#### Iterate and continually improve solutions, services and processes
+
+- Organize training on Agile/DevOps for managers and team leads
```
jaysonmc: Organize training on Agile/DevOps for managers and team leads
=> 
Invest in training on Agile/DevOps for managers and team leads

Organize, I would argue, is too hand wavey. Invest is an action and a request for commitment from the department to allocate resources towards it

presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
```diff
@@ -0,0 +1,141 @@
...
+
+---
+
+### Hackathon
+
+- First of many (1-2 per year)
+- Learning, automation and improvement
+- Digital, agile, DevOps and cloud
+- 2020-02
+
```
jaysonmc: Creation of a DevOps Dojo

presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
smellems: Agile is the only buzzword we have been keeping so far and it's already being promoted in IITB by interop, DTS, us, .. new DG blog yesterday.

But we should have something on reducing lead time..

presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
smellems: Done

presentations/en/Strategy-Continuous-Learning-Automating-Improving.md
smellems: Reducing lead time is one of the Goals. Does that work?

smellems:
@jaysonmc I incorporated most changes from #456. please review.

Keep in mind that these are the strategies for Q4.  I think we should review them quarterly at least.  So if we say we're going to do a Hackathon, it should be early 2020.

Would Dojos fall under learning/events/conferences?

Champions: is that learning or a centre of expertise?

I did'nt want to say use the CATS codes, because I think it would be better to get rid of CATS, at least in its current format (individual employees filling it in).  It could even be part of this strategy, giving people back some time.

jaysonmc:
Long term, yeah. Get rid or automate CATS... for now we have them and people are not continuously improving because there "is no CATS code". So it's an  easy fix, Long term we can still aim to replace it, but no reason not to take a quick win? 

And DevOps Dojo is separate from conferences in that it requires constant commitment (dedicated venue and personnel [even if they rotate]). It's a significant investment compared to just sending folks to conferences.

gcharest:
Some of the content isn't 100% aligned with the referenced document at #456

gcharest:
Approved but would recommend that Working in the open becomes its own set of strategies.

PR #484, opened 2019-12-04
Changes based on feedback (Access to SaaS)

smellems:
Comments to address:

## Issues
I struggle with how DLP figures in to this.  Although I understand we would use it to limit data loss etc… it seems that as the first Item in issues that this is about DLP. Maybe it could be part of the mitigation strategies on how to reduce risk?
We refer to the need to open up access to services to support GOC policy and employee Internet accessibility above.  The Issues don’t feel like they line up well to the purpose.  I like the fact that we are saying we are already doing it but don’t grasp what we are trying to highlight as an issue.

## Considerations
Good do we have any non TBS or digital academy references we could add? (using Google drive, Docs, ..)

## Recomendations
Reference DLP here as a risk mitigation item

## Appendix A
Is there value in adding a column that would explain benefits of use?  When I see chat I say we already have it with Lync.

### DIFF
--- a/_pages/en/web-services-access.md
+++ b/_pages/en/web-services-access.md
@@ -69,7 +69,7 @@ The IT Strategy team will work with teams in IITB to:
 #### Benefits
 
 - Don't be restricted by Slack free plan of having access only to the 10 000 most recent messages
-- Ability to encrypt communications (can be a good practice even for unclassifed information)
+- Ability to encrypt communications (can be a good practice even for unclassified information)
 
 |             | URL           | Notes  |
 | ----------- |:-------------:| ------:|
--- a/_pages/en/web-services-access.md
+++ b/_pages/en/web-services-access.md
@@ -65,7 +65,7 @@ The IT Strategy team will work with teams in IITB to:
 | ----------- |:-------------:| ------:|
 | Hack Chat | https://hack.chat |  |
 | Gitter | https://gitter.im |  |
-| Riot IM | https://riot.im |  |
+| Riot IM | https://riot.im | Client for Matrix homeservers |
 | RocketChat | https://open.rocket.chat |  |
 | Slack | https://slack.com | Websockets blocked |
 
@@ -136,6 +136,8 @@ Note: Inconsistent browser configurations (Websockets, 3rd party cookies blocked
 
 | Name        | URL           | Notes  |
 | ----------- |:-------------:| ------:|
+| Librem One Chat | https://chat.librem.one | Homeserver only, requires client |
+| Matrix.org | https://matrix.org | Homeserver only, requires client |
 | Zulip | https://zulipchat.com |  |
 | ... |  |  |
 
@@ -201,7 +203,7 @@ Note: Inconsistent browser configurations (Websockets, 3rd party cookies blocked
 | Name        | URL           | Notes  |
 | ----------- |:-------------:| ------:|
 | Facebook | https://www.facebook.com |  |
-| Librem One | https://librem.one |  |
+| Librem One | https://social.librem.one |  |
 | Linkedin | https://www.linkedin.com |  |
 | Mastodon | https://mastodon.social |  |
 | Twitter | https://twitter.com |  |
--- a/_pages/en/web-services-access.md
+++ b/_pages/en/web-services-access.md
@@ -21,28 +21,28 @@ This Policy applies only to unclassified assets (data, information, code, ..).
 
 ### Issue
 
-ESDC is currently blocking access to specific popular Web 2.0 collaboration tools like Google Docs, Google Hangouts and OneDrive.
-Even though TBS policy generally recommends open access to these tools, ESDC's rationale in blocking them has been primarily to prevent accidental leak of protected information from employees.
+ESDC is currently blocking access to specific Web 2.0 collaboration tools like Google Docs, Google Hangouts and OneDrive.
+Even though TBS policy generally recommends open access to these tools, ESDC's rationale for blocking them has been primarily to prevent accidental leak of protected information from employees.
 This is a valid concern that must be mitigated with a balanced approach.
 
-The proposed risk mitigation already on the table is to implement better Data Loss Prevention (DLP), including a tool that monitors and enforces classification of files and emails before they can be transmitted on the Internet.
+The proposed risk mitigation would be to augment the current Data Loss Prevention (DLP) by adding a tool that monitors and enforces classification of files and emails before they can be transmitted on the Internet.
 However, as the evaluated options so far can be costly, implementation is not being considered for the short term.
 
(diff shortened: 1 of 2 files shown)


_pages/en/web-services-access.md
```diff
@@ -21,57 +21,99 @@ This Policy applies only to unclassified assets (data, information, code, ..).
...
 
+#### Benefits
+
+- Work on GitHub with SSH keys for better security
+
 |             | URL           | Notes  |
 | ----------- |:-------------:| ------:|
 | Github | https://github.com | Websockets and SSH blocked |
 
 ### Email
```
gcharest: I'm not sure what is implied as an advantage here.

_pages/en/web-services-access.md
```diff
@@ -99,17 +149,12 @@ Should you approve, the IT Strategy team will work with teams in IITB to:
...
-### Collaboration
+### Document Editing
 
 |             | URL           | Notes  |
 | ----------- |:-------------:| ------:|
-| Doodle | https://www.doodle.com |  |
-| Eventbrite | https://www.eventbrite.ca |  |
 | Framapad |  https://framapad.org |  |
 | Office365 (pilot) | https://www.office.com |  |
-| Slido | https://www.sli.do |  |
```
gcharest: Is Slido a Document Editing SaaS? (It's also listed under **Other**)

_pages/en/web-services-access.md
```diff
@@ -21,57 +21,99 @@ This Policy applies only to unclassified assets (data, information, code, ..).
...
 
 ### Considerations
 
 More departments are using SaaS to collaborate and engage with citizens and other public administrations in Canada and the world.
-For example, the latest Open Government Plan and the Service and Digital Target Architecture from TBS were shared as drafts for consultation using Google Docs.
+For example, the latest Open Government Plan and the Service and Digital Target Architecture from TBS were shared as drafts for consultation and comments on Google Docs.
 The GC Digital Academy and Free Agents, as well as other groups, frequently use Google Forms for registration.
 
+In order to move forward, we propose a balanced approach in opening access to additional SaaS tools (See [Appendix A](#appendix-a---saas-to-enable)) to IITB employees in the short term, as a step toward more broader access in ESDC.
+As IITB employees generally don't deal with protected information and are well informed and trained on the management of information, we anticipate the risk to be minimal.
```
gcharest: I think that this is an assumption: DBAs most likely would have access to the data directly. Developers may be able to query DBs with sensitive information/data.

_pages/en/web-services-access.md
```diff
@@ -21,57 +21,99 @@ This Policy applies only to unclassified assets (data, information, code, ..).
 
 ### Issue
 
-To be able to follow the Policy, ESDC needs better Data Loss Prevention (DLP), including a tool that monitors and enforces classification of files and emails before they can be transmitted on the Internet.
+ESDC is currently blocking access to specific Web 2.0 collaboration tools like Google Docs, Google Hangouts and OneDrive.
+Even though TBS policy generally recommends open access to these tools, ESDC's rationale for blocking them has been primarily to prevent accidental leak of protected information from employees.
+This is a valid concern that must be mitigated with a balanced approach.
 
-Technical Architecture has started to include some SaaS in the ESDC Technology Bricks (GitHub, Trello and Slack), but these don't reflect what can be accessed on the ESDC network or what is currently being used by employees (See [Appendix B](#appendix-b---saas-already-unblocked)).
+The proposed risk mitigation would be to augment the current Data Loss Prevention (DLP) by adding a tool that monitors and enforces classification of files and emails before they can be transmitted on the Internet.
```
gcharest: I'd propably recommend we come up with more diverse options for risk mitigation: DLP for email, OK, but more controlled access to DBs so that people who actually have access to sensitive information/data cannot extract large amounts of data, look for open source options to help monitor key systems use, etc.

_pages/en/web-services-access.md
gabrielcossette: SSH keys are more secure than username/password, but I agree this is kinda a moot point if we enable 2FA. How are Websockets used in GitHub (to find benefits in enabling them)?

_pages/en/web-services-access.md
gabrielcossette: You are right, fixed :)

_pages/en/web-services-access.md
gabrielcossette: Indeed, thus why I included the "well informed and trained" part ;)
I assume that DBAs are well trained to deal with protected information and unlikely to just copy/paste them into a random SaaS.

_pages/en/web-services-access.md
gcharest: Argh, my bad, I thought I was tagging the email SaaS service with "unique email" for external IT professionals line :man_facepalming:

_pages/en/web-services-access.md
gcharest: OK, I see what you mean. My point was around trying to frame the "consideration" of access controls. DLP for email is a bit narrow in scope. Technically, there probably should be access audits and automated monitoring as well as a proposed way of mitigating the risks? Say for example, Database Access Monitoring, Application Activity Monitoring, etc. so that we increase our security and confidence in data loss prevention at the source, not at the perimeter level.
Quick read: https://opensource.com/article/19/4/log-analysis-tools

gabrielcossette:
@smellems J'ai fait d'autres ajustements dont entre autres changer "SaaS" pour "Web tools". Je trouve ce terme plus "générique" et accessible. Comme on ne sait pas qui va lire la proposition, je crois que ça serait mieux. Mais je te laisse décider, on peut revenir à SaaS également.

_pages/en/web-services-access.md
```diff
@@ -1,141 +1,185 @@
...
+- Grant access to additional Web tools (see [Appendix A](#appendix-a---web-tools-to-grant-access)) for IITB employees, as a step toward broader open access to the Internet for ESDC employees; and
+- Perform an evaluation of security monitoring tools, active at the source (application and database) and the perimeter (DLP), to mitigate future risks associated with broader open access to the Internet.
 
 ### Next Steps
 
-Should you approve, the IT Strategy team will work with teams in IITB to:
+The IT Strategy team will work with teams in IITB to:
+
+- Ensure the following Web tools (see [Appendix A](#appendix-a---web-tools-to-grant-access)) are functional on the network for IITB employees;
+- Ensure IITB employees' understanding of document classification;
```
gcharest: and that the tools and services are only to be used with unclassified information and data.

gcharest: would also add: work in conjunction with information management to ensure...

_pages/en/web-services-access.md
```diff
@@ -1,141 +1,185 @@
...
+- Review access rights to data and information;
```
gcharest: We will do that ourselves?

Issue #297, opened 2019-10-09
Update profile page with personas for IITB


gcharest:
Update this page
https://github.com/sara-sabr/ITStrategy/blob/master/_pages/en/employee-profiles.md

myp192:
Currently doing some information management with the meeting notes to "build the puzzle" for the personas.

smellems:
Notes from yesterday's meeting.

For each persona, look at
- what does this user use technology for
- what outcomes are they producing
- how many levels away from the client are they

Outcomes should be specific, or be split into more personas.

Map business capabilities to outcomes.

Next: layered - systems of records, systems of differentiation, systems of innovation, ..

PR #512, opened 2019-12-19
Employee profiles

gcharest:
Add content to the WIP Employee Profile page. (English only at the moment).

Few things:

- Initial list doesn't match new list of profiles
- There seems to be empty profiles.


gcharest:
Linked to #297

PR #516, opened 2019-12-19
Elmina changes

ElminaIusifova:
https://www.draw.io/#Uhttps%3A%2F%2Fraw.githubusercontent.com%2Fsara-sabr%2FITStrategy%2FelminaChanges%2Fdrawio%2Fen%2FSoftwareOnboarding-UserFlow.drawio


smellems:
We also already have this draw.io for software onboarding
https://www.draw.io/#Uhttps%3A%2F%2Fraw.githubusercontent.com%2Fsara-sabr%2FITStrategy%2FelminaChanges%2Fdrawio%2Fen%2FSoftwareOnboarding.drawio

ElminaIusifova:
> 1. Number the boxes, so it's easier to reference (e.g. step 4)
> 2. I would also add rejected softwares into the "DML" (catalogue?), so there's a record as to why that was rejected.
> 3. The first check could be to verify whether a requested software is already in the catalogue (rejected or not), as a means to reduce processing
> 4. At step "suggest another version", the flow should loop back at box "checking for approval". This to simplify workflow paths
> 5. Is there a last step missing, after "Software catalogue"? I.e. installation of software on requestor's machine, or unblocking access to web-based SaaS?

ElminaIusifova:
@remyb2canada 
1. Will do
2. We were thinking to add one more tab which will have all rejected requests (which are not approved) and could be filtered by time and name. So DML will have requests only approved 
3. didnt get this point 
4. When the version of the software is denied another version will be suggested to the user, why I am directing him/her  to assessment page  - letting him/her still choose different version as suggested version might not be the user choice 
5. Right! Missing this step

ElminaIusifova:
created a new PR request

PR #511, opened 2019-12-19
Profiles

gcharest:
Created a data folder where we can add our professional profiles.

Next step will be to create the cards pulling the data from them.

PR #517, opened 2019-12-19
Add data for profile cards

gcharest:
Too braindead to fix #511; new PR with data files only, haven't touched the employee profile page.


_data/profiles/remy-bernard.yml
```diff
@@ -0,0 +1,14 @@
+---
+name: Rémy Bernard
+role:
+  en: Genius
```
gcharest: I have some doubts about the validity of the supposed facts listed on this profile's data... :thinking:

remyb2canada: I will support it by an Expert opinion:

It is true

😄

gcharest: Wait, you'll bring in an impartial 3rd party?

PR #524, opened 2019-12-30
Software onboarding  process

ElminaIusifova:
https://www.draw.io/#Uhttps%3A%2F%2Fraw.githubusercontent.com%2Fsara-sabr%2FITStrategy%2FelminaChangesV2%2Fdrawio%2Fen%2FSoftOnboarding.drawio


ElminaIusifova:
#516

gabrielcossette:
I pushed some proposed updates to the diagram. Let me know what you think. If they make sense, some renumbering will be needed.

jaysonmc:
I dont think between number 5 and 6 we need to suggest a different version. 

If they want a new version of the software, then assess it. 
If it fails, suggest the version that passed
If it succeeds, then inform then it was approved

ElminaIusifova:
@jaysonmc this is exactly what we are suggesting.  If it fails, suggest the version that passed, means the version that exists in DML.

Issue #467, opened 2019-11-29
Merge pending UI changes


jaysonmc:
Branch: https://github.com/sara-sabr/ITStrategy/tree/newwebpage

gcharest:
Moved changes to branch [**ui-changes**](https://github.com/sara-sabr/ITStrategy/tree/ui-changes)

PR #532, opened 2020-01-07
Ui changes

gcharest:
- Updated the Splash screen to be more alive (changed Jumbotron to custom display)
- Lightened up the overall theme (more white and contrast)
- Added dynamic drop-down menus to the following sections:
  - About us
  - Ready for use documents (aka Our strategies)
- Renamed Working in the Open page as it was way too long.

A few things to work out in upcoming PRs:
- Navigation structure (should be a different PR altogether, requires better information architecture discussion)
- Trigger the collapse of the menu at higher resolution (currently at mobile) so that ipad display looks good as well


gcharest:
Fixes #467

smellems:
I like that this doesn't add too many new Javascript and CSS files and is more improving what we have with bootstrap. 😄

gabrielcossette:
I reduced the margins of the header section (as it was before). Let me know if you prefer to keep larger one.

gabrielcossette:
If we are on a page without a FR translation, the "Our Strategies" sub-menu is too large (see image):
image

gabrielcossette:
2 suggestions:
1. Display the SABR logo top left instead of "IT STRATEGY" (duplicate from header section)?
2. Add change of color when hovering items in menu?

gabrielcossette:
Forgot to mention above that if you agree with my suggestions, I can help with the work!

gcharest:
@smellems Good catch on the image, will double check if actually in used and if so, will need to compress.

> 2 suggestions:
> 
>     1. Display the SABR logo top left instead of "IT STRATEGY" (duplicate from header section)?
> 
>     2. Add change of color when hovering items in menu?

@gabrielcossette I like the logo idea! Regarding hover in menu: 100%

gcharest:
> If we are on a page without a FR translation, the "Our Strategies" sub-menu is too large (see image):
> image

You mean the white space at the right?

gabrielcossette:
> 
> 
> > If we are on a page without a FR translation, the "Our Strategies" sub-menu is too large (see image):
> > image
> 
> You mean the white space at the right?

I mean, the title is cut out. We can't see the full title of this particular Strategy.

gcharest:
> > > If we are on a page without a FR translation, the "Our Strategies" sub-menu is too large (see image):
> > > image
> > 
> > 
> > You mean the white space at the right?
> 
> I mean, the title is cut out. We can't see the full title of this particular Strategy.

Hmmm... Need to better think those links then. Either the page title is too long itself or we need it wrapped?

gcharest:
@gabrielcossette Should we merge now or address the dropdown points first?

Issue #560, opened 2020-01-10
No spellcheck for links and images


gabrielcossette:
Currently, cspell is checking spelling in links and images, leading to unknown words. We should add an exception.

Related to discussion in #553

PR #568, opened 2020-01-12
Add regexp cspell exclude on markdown links

gcharest:
- Removed "lamélioration" from exceptions
- Added regexp exception to test


cspell.json
```diff
@@ -286,8 +286,7 @@
     "minusxl",
     "comparecurves",
     "areaundercurve",
-    "integralareaundercurve",
-    "lamélioration"
+    "integralareaundercurve"
```
smellems: I thinks most of these were added because of file names in links.

smellems:
Fixes #560

PR #579, opened 2020-01-14
Software onboardin interactive prototype V1

ElminaIusifova:
Link to prototype you can find here:

https://xd.adobe.com/view/36efe61b-5b19-4f00-43af-f97bd9c2db65-40de/


ElminaIusifova:
**In Software Onboarding (v1) we have 4 scenarios:**
Start with Welcome page (Menu-> Make a request -> New request);
**Scenario 1.**  User is requesting a software that exists in DML/Software Catalogue. 
**Scenario 2.** User is requesting  a software which is unable to be installed for one of the following reasons - license compliance, security check, accessibility (doesn't apply for CML tool). `Click Red REJ button`
**Scenario 3.**  User is requesting a software that passes license compliance, security check, accessibility but doesn't exist in DML/Software catalog. The user must add the software to Catalogue manually. `Click Not in DML purple button`
**Scenario 4.** User is requesting a software that exists in DML/Software catalog but with a different version that doesn't pass license compliance, security check, accessibility. Another version is suggested for a user to accept or deny. `Click Green suggest button`

ElminaIusifova:
to review user workflow see issue #524

smellems:
The example used : Ubuntu is not a command line tool.

I would change it to markdownlint-cli version 0.21.0 (MIT licence)
https://www.npmjs.com/package/markdownlint-cli

ElminaIusifova:
@smellems true, will fix that

gabrielcossette:
Looks awesome, great work!

Some suggestions:

**Scenarios**
* **3** - It's rather that the desktop team will subsequently add the software to the Catalogue and the user is allowed to install it himself.
* **4** - Different version is not checked for compliance right away. The already approved version is offered and if user refuses, then we go through compliance checks and the other scenarios are triggered.

**General**
* Clicking "Next" or "NOT IN DML" should bring up the same "Checking for approval" screen no (for consistency)?
* The screen after clicking "Next" should not offer to open Application Catalogue because the newly approved software needs to be packaged to be in the Catalogue. Instead mention that the user is now allowed to install the software himself and that the desktop team will look at adding it to the Catalogue.
* The icon for Accessibility check shouldn't be like a "fail" because the check it is actually out-of-scope. It should be an icon less "negative", maybe with an asterisk "*" with a mention below that it is out-of-scope.
* No need for a dropdown in the menu "Make a request". Just make "New Request" a top-level item.

**Suggest**
* Maybe replace with something like "A different version of your requested software has already been approved. Would you like to install version 2.1.2 instead?" 
* If select "No", bring up the screen "Checking for approval..." (instead of "Thank you").

**Not in DML**
* Could add another screen to acknowledge that software can now be installed, will be added to DML and desktop will look at adding it to Catalogue.

gcharest:
> Looks awesome, great work!
> 
> Some suggestions:
> 
> **Scenarios**
> 
>     * **3** - It's rather that the desktop team will subsequently add the software to the Catalogue and the user is allowed to install it himself.
> 
>     * **4** - Different version is not checked for compliance right away. The already approved version is offered and if user refuses, then we go through compliance checks and the other scenarios are triggered.
> 
> 
> **General**
> 
>     * Clicking "Next" or "NOT IN DML" should bring up the same "Checking for approval" screen no (for consistency)?
> 
>     * The screen after clicking "Next" should not offer to open Application Catalogue because the newly approved software needs to be packaged to be in the Catalogue. Instead mention that the user is now allowed to install the software himself and that the desktop team will look at adding it to the Catalogue.
> 
>     * The icon for Accessibility check shouldn't be like a "fail" because the check it is actually out-of-scope. It should be an icon less "negative", maybe with an asterisk "*" with a mention below that it is out-of-scope.
> 
>     * No need for a dropdown in the menu "Make a request". Just make "New Request" a top-level item.
> 
> 
> **Suggest**
> 
>     * Maybe replace with something like "A different version of your requested software has already been approved. Would you like to install version 2.1.2 instead?"
> 
>     * If select "No", bring up the screen "Checking for approval..." (instead of "Thank you").
> 
> 
> **Not in DML**
> 
>     * Could add another screen to acknowledge that software can now be installed, will be added to DML and desktop will look at adding it to Catalogue.

I think @gabrielcossette 's observations are very good. I don't have more to add other than we probably should make sure that we don't use any acronyms (or at least limit their use and include an hover "definition" bubble - if that's accessible)

gcharest:
Just my personal feedback: I'd rather avoid lightbox as part of a workflow? Things should be done in a single page as much as possible or at least have a very thorough set of testing to avoid moving things around too much?

ElminaIusifova:
The link to the prototype with new updates has been changed. Waiting for feedback

gcharest:
> Is this only the link?
> merge 21 commits into elminaChangesV2 from master
> Blocking so we don't merge this. but lets review link

Wow, totally did not see your comment here 🤦‍♂

gcharest:
@ElminaIusifova When we get to the "Not in DML", where is the list of categories for the drowdown supposed to ask for the "licensing supply arrangement" coming from? Is that the current process (ish)?

ElminaIusifova:
@gcharest  according to @gabrielcossette  yes and this is where it comes from 
https://www.tpsgc-pwgsc.gc.ca/app-acq/cral-sarc/lgcl-ctgr-eng.html

smellems:
Page 1
Is this only for command line tools?  I thought it was and it's in the first question asked.

This generally means packages that are installable through NPM (Node is already available on Desktops) or Python, Ruby, Composer, ..  The example, MarkdownLint-cli is an NPM package.

Page 2
Options for "what is this tool for" should be:
- Production (used for building or as part of deployed application or service)
- CI and CD (for automated testing and deployments only)
- Experimentation (playing around, not official use)

For experimentation, it's already allowed as long as the technology/framework is available.  For example, Node is available and you can install any package you want to play around with.

Page 3-4
We could have the licence and security checks as part of the process, to help provide a way to test packages before starting to use them.  But it's just as important that teams include these checks as part of their CI tests.

Maybe CI/CD and Experimentation shouldn't require any approvals.

Page 11-12
What is the DML?  Is it for all software on only desktop stuff?
For the example used, the Category should be Development environments software.

Would all command line tools be added to DML?  Maybe only those used for Production.
Is Node in the DML?  Meaning you can install all NPM packages?

Page 13
Desktop team?  The desktop team would approve Ruby, Python, R, Composer or other frameworks that allow installation of packages..

PR #604, opened 2020-01-20
Update Q1 strategy actions guidance

smellems:
As discussed I removed Automation from the action for the Q1 strategy.

I also updated the list of guidance based on discussion.  I changed it back to a list from a table to make it easier to comment and edit.  We can always go back to a table for better presentation later.

We should rename the file name and permalink, but that would break links we shared already

### DIFF
--- a/_pages/en/strategy-learning-automating-improving.md
+++ b/_pages/en/strategy-learning-automating-improving.md
@@ -43,7 +43,8 @@ All IITB teams must:
 
 ##### Improve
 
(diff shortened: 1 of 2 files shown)


myp192:
I added two bullets in the improve sections: 
- Understand teams performance levels and set realistic goals
- Identify gaps in your daily work performance and establish improvement priorities

I think I forgot to do something because it's not showing my changes in the file (my bad)

PR #608, opened 2020-01-21
Localstorage to store user language preference and gemfile 

Moro-Code:
- Use localstorage to store users language preference so that they only see the splash page once. If valid language exists in localstorage user will be routed to respective index page 
- Added gemfile to be able to run ```bundle install``` on local machine.


gabrielcossette:
Great idea! Added a fix in Gemfile and Jekyll config.
However, if we later switch to the other language (top right), should we replace the langauge preference? That's not the case at the moment.

Moro-Code:
Alright will add this first

Moro-Code:
@gabrielcossette completed, please review

Issue #649, opened 2020-01-28
Accessibility: Link Name for Home nav link


gcharest:
Ran a quick accessibility test with Accessibility Insights for Web on Chrome and found the following failed test:

Path | .navbar-brand
-- | --

`<a class="navbar-brand" href="/ITStrategy/accueil.html">`

Fix the following:

- Element is in tab order and does not have accessible text

Fix ONE of the following:

- Element does not have text that is visible to screen readers
- aria-label attribute does not exist or is empty
- aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
- Element's default semantics were not overridden with role="presentation"
- Element's default semantics were not overridden with role="none"

Moro-Code:
On it... thanks for the check

Moro-Code:
resolved by pull  #656

PR #656, opened 2020-01-30
Fix easy accessibility errors found with WAVE

smellems:
Tested Website using https://addons.mozilla.org/en-CA/firefox/addon/wave-accessibility-tool/

Fix error with the home logo not having a label. And removed extra HTML `</li></a>` from the same line

Applied the style for links in paragraph to all links in main content.  This does not affect the header or footer.  But the notice is part of main and the PR change the link color from the default alert color to blue.


Moro-Code:
Is this similar to #652 ?

smellems:
Closes #649

PR #719, opened 2020-01-30
Contlearn

ElminaIusifova:
I tried to visualize what we discussed and drew on the board at the meeting. Please comment and provide feedback


ElminaIusifova:
This PR is disabled, I will create a new one

PR #807, opened 2020-02-20
Add references to Continuous Learning Strategy

gabrielcossette:
Incl. minor edits

Feel free to add more references!


smellems:
Is this a good place to add all these references?  It make the page much longer and seems more complex.  The idea was to keep it simple and to the point.

We should have links to Accelerate and DORA with quick explanations in Goals and Actions.
Also a "why 20%" section under Actions with a paragraph explanation and some links.

I guess it depends what this page is for.  I thinks it's more to communicate what you should do for Q1, and not so much why.

We will have to explain why in more details when we seek endorsment and some of these quotes and stats would be good for that. Maybe a presentation

PR #820, opened 2020-02-26
Moving enabling directorate page into archives

jaysonmc:


jaysonmc:
The following link is dead
https://canada-ca.github.io/PCTF-CCP/overview/pctf-overview.html
Now it seems to be ..
https://canada-ca.github.io/PCTF-CCP/
It is used in the ITVision.md file

Issue #850, opened 2020-03-05
Continuous Learning Strategy


jaysonmc:
Recommend which CATS codes to use for which actions outlined in the guidance 

Ex: Grassroots has it`s own CATS code. Use that one. It will be mapped behind the scenes. 
Taking online courses, that could be mapped to training

Automating repetitive tasks, that would go under Continuous Improvement, for example

jaysonmc:
#853

jaysonmc:
PR needs to be resolved (currently breaking build), then we can merge and close 

#853

PR #853, opened 2020-03-06
Continuous improvement update

jaysonmc:
If anyone else knows of any other CATS codes that could apply to what we have listed, I think that would help folks

Linked to issue #850


jaysonmc:
Great, thanks @gabrielcossette 
I propose we wait to merge until tomorrow so @RachelMuston can take a look too

gabrielcossette:
Also added CATS for **IM Clean-up**, which fits under continuous improvement I believe.

PR #868, opened 2020-03-16
Add Strategy and Analysis sections for Documents page

### LINKED ISSUE #844: Create Strategy Category on Site

smellems:
Closes #844 

This replaces "Ready for use" with a "Strategies" category on the Documents page.

I also added a "Done Enough" category.

I archived the Strategies and Actions page.  are there other pages we could move from "Work in Progress" to "Archive"?

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,7 +19,7 @@ description:
   fr: Travailler ouvertement! Nos travaux en cours, liste de documents de références et présentations.
 sectionsList:
   en: ["About us", "Strategies", "Done Enough", "Work In Progress", "Archives"]
-  fr: ["À propos de nous", "Stratégies", "Assez fait", "Travaux en cours", "Archives"]
+  fr: ["À propos de nous", "Stratégies", "Assez fait", "Travail en cours", "Archives"]
 docsTitle:
   en: Documents
   fr: Documents
--- a/_config.yml
+++ b/_config.yml
@@ -18,8 +18,8 @@ description:
   en: Working in the open! Our work in progress, list of reference material and presentations.
   fr: Travailler ouvertement! Nos travaux en cours, liste de documents de références et présentations.
 sectionsList:
-  en: ["About us", "Strategies", "Work In Progress", "Archives"]
-  fr: ["À propos de nous", "Stratégies", "Travaux en cours", "Archives"]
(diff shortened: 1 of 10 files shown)


jaysonmc:
@smellems 
Perhaps we should put Strategies on the right hand side and have a subsection for quarterly strategies (as that list will be long over time?)

smellems:
I think it should be a category for things that are "done" and were used somewhere or we think there good and don't want to archive them.  I don't think "North Star" or "Direction" describe that.  Maybe somethings else..

I'll move HDLC

jaysonmc:
Done is a little vague for users though, or even for us.

It is done, but why? I would suggest that the category should also say something about the documents use

This is a list of done documents
vs
List of strategies
List of blogs
List of About Us
List of References

What do you think?

ElminaIusifova:
Why not put all strategies (A Strategy to transition IITB to an ESDC partner, 
Build vs Buy Strategy)  under the Strategies category?

jaysonmc:
@ElminaIusifova 
Would work fine for now, but the quarterly strategies will get long and dwarf all the other (larger) ones.

smellems:
Latest - Analysis would be between About us and Work in progress

image

I agreed we could work on the presentation, but let do that in a separate PR

smellems:
@ElminaIusifova We could move them to Strategy, but I don't know if there ready to be moved up or if they should be archived

smellems:
Closes #887

ElminaIusifova:
Would also make Quarterly Strategies as a submenu of Strategies

ElminaIusifova:
Asking as a user: What "Work in Progress" and "Archives" sections means?

smellems:
I would say, "things we are working on" and "things we don't use anymore".  Maybe archives could just be removed.  Those pages would stay on Github

ElminaIusifova:
A strategy map is a key to all our strategies, maybe we can move it as a separate section (button) on the main page or link it to all our strategy pages only

ElminaIusifova:
Why would we keep something that we don't use anymore on our website? unless it is something that can be informative and useful for a user, in this case, we can move it under a different tab.  Also, it depends on the purpose of the website of course, What message are we trying to deliver?

PR #871, opened 2020-03-16
Q1 strategy alignment with HR

ElminaIusifova:


_pages/en/strategy-learning-automating-improving.md
```diff
@@ -35,6 +35,10 @@ For more books, reports and articles supporting this strategy, you can consult o
   - [Mandatory Procedures for Enterprise Architecture (EA) Assessment](https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32602)
   - [Mandatory Procedures on Application Programming Interfaces (API)](https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32604)
   - [Acceptable Network and Device Use](https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32605)
+ - [Transformational aspects of HR Learning Priorities](http://iservice.prv/eng/college/learning_priorities_calendar.shtml)
+  - [Digital Mindset](http://iservice.prv/eng/college/docs/learning_priorities.pdf)
+  - [Digital technology](http://iservice.prv/eng/college/docs/learning_priorities.pdf)
+  - [Foundational skills: “Management/Leadership”](http://iservice.prv/eng/college/docs/learning_priorities.pdf)
```
smellems: The header is for GC IT Direction under Goals. Should we add one for ESDC IT Direction?  If so we should have the IT plan and the way forward there as well.

Maybe it would be better to add guidance under Learn with the links to departmental learning priorities and calendars.

The last 3 links are the same.

ElminaIusifova: as for links: will double-check them when access to network will be available

_pages/en/strategy-learning-automating-improving.md
```diff
@@ -35,6 +35,10 @@ For more books, reports and articles supporting this strategy, you can consult o
   - [Mandatory Procedures for Enterprise Architecture (EA) Assessment](https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32602)
   - [Mandatory Procedures on Application Programming Interfaces (API)](https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32604)
   - [Acceptable Network and Device Use](https://www.tbs-sct.gc.ca/pol/doc-eng.aspx?id=32605)
+- [ESDC Learning Priorities](http://iservice.prv/eng/college/learning_priorities_calendar.shtml)
+  - Digital Mindset
+  - Digital technology
+  - Foundational skills: “Management/Leadership”
```
smellems: The header is for GC IT Direction under Goals. Should we add one for ESDC IT Direction? If so we should have the IT plan and the way forward there as well.

I think it would be better to add guidance under Learn with the links to departmental learning priorities and calendars.

jaysonmc: Sure, that works too.

PR #881, opened 2020-03-17
Adding language learning resources & reformatting learning section

jaysonmc:


remyb2canada:
This is starting to be pretty big. I worry the document ends up being a google search result with so much information that you need a search tool to go through it :)

Perhaps we have to scope its content as such:

1. ESDC And GoC Alignment materials (must be materials that can be used by IT as instruments to demonstrate alignment. I.e. whatever IT does, it must be within the range of those materials)
2. Materials that the IT Strategy team uses. If we don't use it, we don't list it. These include worthwhile book materials, online information, tutorials, software (if they are not worthwhile, we don't list them perhaps?)
3. Crowd Sourcing info. This would relate to [this milestone](https://github.com/sara-sabr/ITStrategy/milestone/45) that @Moro-Code is working on so the community has the ability to add on to it

jaysonmc:
Aye, it is getting sizeable. But we were asked for a page with the content we've added (tools folks can use and educational stuff folks can do awhile off network)

After this merge, I agree. I'll create an issue for splitting this out into numerous documents. Maybe even a new section on the site?

PR #899, opened 2020-03-24
Readd About us to menu

smellems:
About us disapeared from the menu (became Vision) during the reorg of sections on the site.

PR adds it back.

This is what it looks like
image


ElminaIusifova:
Not only About Us page, but there are also few which disappeared after the last merge. I think you  can add it for now with this PR and now I am working or restructuring the navigation menu which can be implemented later on

PR #1014, opened 2020-05-25
Change to Website based on ROT analysis

smellems:
- Remove notice and add modal message with link in footer
- Archive some pages (How to contribute, Q1 and Strategy Map)
- Remove Mandate from every page
- Remove some words and links from Home page
- Change green to blue (info alert)
- Add Archive section to list archived pages

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -18,8 +18,8 @@ description:
   en: Working in the open! Our work in progress, list of reference material and presentations.
   fr: Travailler ouvertement! Nos travaux en cours, liste de documents de références et présentations.
 sectionsList:
-  en: ["Vision", "Strategies", "Quarterly Strategies", "About us", "Analysis", "Work In Progress"]
-  fr: ["Vision", "Stratégies", "Stratégies trimestrielles", "À propos de nous", "Analyses", "Travail en cours"]
+  en: ["Vision", "Strategies", "Analysis", "Work In Progress"]
+  fr: ["Vision", "Stratégies", "Analyses", "Travail en cours"]
 docsTitle:
   en: Documents
   fr: Documents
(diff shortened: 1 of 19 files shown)


_layouts/default.html
```diff
@@ -10,7 +10,6 @@
     <main role="main" class="container">
       <section class="jumbotron d-print-none">
         <h1 class="display-3">{{ site.title[page.lang] }}</h1>
-        <h2>{{ site.subtitle[page.lang] }}</h2>
```
remyb2canada, line 13: I would keep the mandate statement on every page. It's what everything we do ties to and is part of the propaganda machine

_config.yml
```diff
@@ -77,9 +77,12 @@ lastModified:
 blogPosts:
   en: Blog
   fr: Blogue
-notice:
```
remyb2canada, line 80: We need to keep this notice. We don't want to mislead the public. Only when the CIO approves our work does it become official.
I'd even say this notice needs to be at the top of every page.
The notice should say that once something is "approved" or "official" that it is found _add link to where it should be (like Sharepoint or the official ESDC public page)_

_layouts/default.html
ElminaIusifova: @remyb2canada which part of Mandate statement you would keep on every page? If we have our mandate on every page then we will have to remove <Mandate> submenu , it will be confusing

ElminaIusifova: Eventually, Mandate can go to our home(landing page)

remyb2canada: the mission statement: 
> Enable the strategic value of IT within ESDC by reducing its risks and increasing business agility

_config.yml
smellems: The notice is still there.  In the footer next to View on Github, the Notice link opens a modal (popup) with the notice.

Is that good?

PR #1025, opened 2020-05-27
Blog project fix excerpt

gcharest:
- Fixed Excerpt (whole post was in the excerpt of the Blog list)
- Added a few return of line where multiple sentences were on the same line.


gcharest:
**Previously:**

image

**Fix:**
image

PR #1055, opened 2020-06-17
Supporting documents for Target IT solution delivery model strategy

remyb2canada:
Relates to PR #1052 

The PR #1052 (update to the Target IT Solution Delivery Model Strategy) will link to 2 supporting documents:
- one that explains the intent behind the strategy (the problem statements it intends to solve)
- one that lists the areas of concerns raised during peer reviews (for transparency and exposing where we need help in solving)


_pages/en/target-solution-delivery-model-strategy-areas-of-concerns.md
```diff
@@ -0,0 +1,416 @@
...
+- Enterprise Architecture
+- Chief Data Office
+
+#### 2.1.3. What is being done about this concern <!-- omit in toc -->
+
+*Coordinating activities*
+
+If a given project requires a given competency, then that competency should exist within the team.
+As a result, if data science is a competency that is required for a given value stream to deliver value to its users, then data science capabilities should be embedded within the team.
+As for how those services are delivered, the technologies should be in place through internal such that the team is able to manage the software required to deliver the data analysis capabilities they need.
```
gcharest, line 98: Worked on the format and overall typos. I think I got most of them but I'm not 100% what you were referring to here:

> the technologies should be in place through **internal** such that

remyb2canada: oh oh @jaysonmc ? :)

PR #1073, opened 2020-06-25
Target Solution Delivery Model - areas of concerns mods

remyb2canada:
2 modifications:

1- modified concern around compliance with Enterprise Data Model following conversation with DAS
2- added new concern on the use of the term "Development Value Stream" that can conflict with Business Architecture's "Value Stream"


gcharest:
Merging. The build error is related to SurveyMonkey being under maintenance. Temporary exception introduced in Master for the time being.

PR #1082, opened 2020-07-03
Buy build datatable

gcharest:
Automating table generation from site data.


jaysonmc:
Neat idea! Im approving because it works (so 'good enough'), but haven't thoroughly reviewed the content yet. Will do this morning and add any thoughts 

Thought could be merged as is, given it's WIP any who

gcharest:
Content doesn't change much and is still in alpha version. Most content revision will be dealt with in [PR 1083](https://github.com/sara-sabr/ITStrategy/pull/1082).

The table generation is using [Datatables](https://datatables.net/) which renders the tables with JS. Will need additional testing for accessibility in tables.

PR #1088, opened 2020-07-07
build vs buy scenarios initial commit

Moro-Code:


gcharest:
> Great work!
> 
> Should we not also provide an example of a scenario for something like Word, or accounting software? We changed the model to IT Solution, and this break down makes sense if you approach it from a build perspective (which works for most questions). But in the case of Word, I imagine having the 'buying' of Word in mind and looking at the 'breaking down into services' and thinking ... database? Backend? Front end? Im just talking about Microsoft Word.
> 
> Perhaps some guidance on how our framework would apply to 'common' decisions, like Word or an email client may be helpful.

Yes, totally can add a "Word processing" scenario. Do we need it as a first iteration?

Issue #882, opened 2020-03-17
Organize References page


jaysonmc:
Use Guillaumes table component to organize the content on the page (similar to what is used [here](https://sara-sabr.github.io/ITStrategy/strategy-build-vs-buy.html) for the Coherent set of Actions section)

gcharest:
Also addresses issue #713

PR #1165, opened 2020-08-07
Reference table

gcharest:
Move manual entry for first two list of references into data tables.


gcharest:
Main changes to this PR:

**`references.yml`**

- Added data for first two tables for the [References](https://sara-sabr.github.io/ITStrategy/references-en.html) page, French and English

**`footer.html`**

- Removed loop call of `datatable.js`

**`table.html`**

- Commented the comment with liquid tags for Jekyll build to ignore (HTML-style comments are included in the site during the build so you could see the comment in the HTML of the page :laughing: )
- Added a condition to add `class="display"` when both conditions are true:
  - The page's front matter contains the table's ID in the datatable-ids variable
  - The data for the table must contain a header, otherwise datatables will not be able to properly load.

**`en/references.md`** and **`fr/references.md`**

- Changed manual tables with the data tables.

smellems:
Part of #882

PR #1174, opened 2020-08-14
Changes to guiding policy and add some actions - roadmap

smellems:


_data/tables/transform-high-performing-org.yml
```diff
@@ -39,22 +44,60 @@
...
         list: false
         fr: 
-        en: Establish communication plan and change management for continuous improvement in (and around) IITB
+        en: Develop a self assessement tool based on capabilities of high performing organizations and PwC recommendations
     - col:
       data:
         list: false
         fr: 
-        en: 
+        en: To help teams understand the current condition and set measurable targets, the self assessment tool will guide teams toward ways to improve based on an evaluation of diferent capabilities linked to high performing organizations.
```
jaysonmc, line 52: Id change language from "to help" to 

Team managers set measureable targets for their teams to improve upon, which are tied to the performance improvements of their management (terminating at the CIO). 

Id vote for more definitive actions. "to help" is a bit hand wavey ... we could still help and they could still not properly define their metrics. I would just state "define their metrics" as the action

smellems: That is the description.  the action here is : "Develop a self assessement tool based on capabilities of high performing organizations and PwC recommendations"

_data/tables/transform-high-performing-org.yml
```diff
@@ -39,22 +44,60 @@
...
+    - col:
+      data:
+        list: false
+        fr: Gouvernance
+        en: Governance
+    - col:
+      data:
+        list: false
+        fr: 
+        en: Define how DGs will guide their teams to achieve the goals of the Guiding Policy
```
jaysonmc, line 80: Define how DGs will guide their teams to achieve the goals of the Guiding Policy
==>
DGs shall define their own continuous improvement metrics (tied to that of the CIO)

_data/tables/transform-high-performing-org.yml
```diff
@@ -39,22 +44,60 @@
...
+    - col:
+      data:
+        list: false
+        fr: 
+        en: The goals of the Guiding Policy is to move towards a Generative (Westrum) culture; to reduce lead time, change failure rate and time to restore service; while increassing deployement frequency. DGs will need to identify how their teams can contribute to the improvement of these metrics, by defining and improving their own metrics.
```
jaysonmc, line 85: I think we can remove this one. It is a desired outcome, not so much an action in and of itself

_data/tables/transform-high-performing-org.yml
```diff
@@ -67,22 +110,27 @@
...
         list: false
         fr: 
-        en: Develop a self assessement tool based on capabilities of high performing organizations and PwC recommendations
+        en: Establish communication plan and change management for continuous improvement in (and around) IITB
     - col:
       data:
         list: false
         fr: 
-        en: 
+        en: Communicating the intended outcome to IITB employees will be very important for this transformation to succeed.
```
jaysonmc, line 118: Communicating the intended outcome to IITB employees will be very important for this transformation to succeed.
==>
Communicate the continous improvement framework to all employees within IITB

_data/tables/transform-high-performing-org.yml
```diff
@@ -100,7 +148,7 @@
       data:
         list: false
         fr: 
-        en: 
+        en: Grassroot initiative provides IITB employees with a way to get approval to experiment with new way of working or improvements. PoC is for teams outside of IITB to request that IITB investigate possible solutions before it submits an official project.
```
jaysonmc, line 151: ?? Is this a comment or an action

_data/tables/transform-high-performing-org.yml
```diff
@@ -188,7 +251,7 @@
       data:
         list: false
         fr: 
-        en: Create system to store and analyze key metrics from teams
+        en: Analyze progress and key metrics from teams to update this stratgegy quarterly
```
jaysonmc, line 254: Analyze progress and key metrics from teams to update this stratgegy quarterly
==>
Reassess metrics and their respective impacts on a bi-annual basis

I mean, if metric y was meant to improve metric x, and y improved but x did not, when reassess if y is actually having the intended consequences. Change metric, or investigate why the increase in y did not have the intended impact

_data/tables/transform-high-performing-org.yml
smellems: There description for the actions.

_data/tables/transform-high-performing-org.yml
smellems: This action is the the IT strategy team to look at progress and update this strategy.  Teams should be continuously reassessing as per the Guiding policy: "IITB teams seek to understand the current condition, set measurable targets for a future date and experiment with ways to achieve these targets until the future date is reached.
When the results have been captured and new targets are set, repeat the process."

_data/tables/transform-high-performing-org.yml
smellems: The guiding policy says: "IITB teams seek to understand the current condition, set measurable targets for a future date and experiment with ways to achieve these targets until the future date is reached."  That includes DGs.

This was an actions for CIOEC to figure out how to move forward with the strategy

jaysonmc: Isnt that what we're supposed to be doing? Figuring out how to move forward with the strategy?

_data/tables/transform-high-performing-org.yml
jaysonmc: Aye, the intent was to make the actions more like actions and less hand wavey. Analyzing on its own isnt a useful action, its the outcome of said analysis. The outcome of this analysis should be an adjustment of key metrics. So the action should be the action instead of the thing we need to do before the action

jaysonmc (approved):
Removing my request for review

PR #1175, opened 2020-08-14
Move translated expressions out of config to site.data

gcharest:
Moved translated expressions to a separate file leveraging site.data in previous PR.

Now updated all paths to use the site.data file instead of the config file.

This doesn't change anything with regards to the building of the site but allows for dynamic change of terms when coding.

Example:

The team name changes from `Information Technology Strategy Team` to `Information Technology Strategy of Awesomeness Team`

When running the site locally to debug, any change to the config file requires a restart of your local server. By using site.data, don't need to restart the server. Magic...


jaysonmc:
Could just be me, but doesnt seem to be working with WSL 

I went into general.yml, head.title.en and changed it to 

en: Information Technology Strategy Team 2

And nothing changed. I then stopped and restarted the server, and it refreshed 

I know WSL has problems with live updated, but that's generally only if you are running against the Windows ifle system (mnt/d/whatever), but I copied all my files over to a native linux drive (/home/whatever), and I got the same behaviour in both places (no automatic refresh)

PR #1227, opened 2020-09-23
Update/review questions for Technical and Process capabilities

smellems:
- Standard format for sections, questions, scoring and recommendations
- Types of questions listed at the top (3 types)
- Some questions in Process, I think are more based on the how to improve the capabilities.  We could keep those for recommendations.


myp192 (approved):
Agreed that the process section will be easier to measure for the assessment :D

PR #1254, opened 2020-09-30
Change gcdevex to micro procurement

RachelMuston:
This PR includes updates to the page which was called GC Developers Exchange.  I've renamed the page to micro-procurement pilot and updated some of the content to reflect the pilot name.  I also added links to the micro-procurement blog post and the micro-procurement deck.

### DIFF
--- a/_pages/en/Micro-procurement.md
+++ b/_pages/en/Micro-procurement.md
@@ -8,6 +8,10 @@ sections: Work In Progress
 permalink: /Micro-procurement.html
 ---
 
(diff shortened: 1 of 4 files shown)


gcharest (approved):
Great realignment! :+1:

smellems:
We should also update French page.

RachelMuston:
> We should also update French page.

Good point @smellems!  Thanks for reminding me. :) I've made the updates to the french page.  If someone could check my translations, I'd be grateful!

smellems (approved):

PR #1287, opened 2020-10-02
updates to the micro-procurement deck

RachelMuston:
includes: re-orienting to micro-procurement pilot rather than 'gcdevex', and ensuring next steps reflect that first we need approval then start the pilot.


RachelMuston:
Linked to issue #1250

smellems (approved):

Issue #730, opened 2020-01-31
Images need more descriptive alt text on page /2019/11/13/case-continuous-improvement-deeper-dive.html


Moro-Code:
**Description**
Alt text for images needs to be more descriptive. 

**Device Information**  
*Device*:  
*Operating System (incl. version)*:  
*Browser (incl. version)*:  
*Additional Details*:

**Element(s) Information**  
*Page*:  https://sara-sabr.github.io/ITStrategy/2019/11/13/case-continuous-improvement-deeper-dive.html
*CSS Selector(s) (if possible and applicable)*:  
```body > main > p:nth-child(7) > img```
```body > main > p:nth-child(9) > img```
*Additional Identifier(s) (if possible and applicable)*:  
*Additional Details*:  

**Screenshot(s) (if possible)**

This has an alt tag of "Area between curves"

<img width="820" alt="Screen Shot 2020-01-31 at 9 43 43 AM" src="https://user-images.githubusercontent.com/26819992/73548129-3d7c2b00-440e-11ea-9468-ccdf8b5fc8b8.png">

This has an alt tag of "Area under curve function"

<img width="530" alt="Screen Shot 2020-01-31 at 9 43 49 AM" src="https://user-images.githubusercontent.com/26819992/73548202-5c7abd00-440e-11ea-8183-61979af32aa8.png">

**Additional information or thoughts**

results from scan #702

PR #1303, opened 2020-10-13
Description to alt text on image tags

ElminaIusifova:


_posts/en/2019-10-15-case-continuous-improvement-deeper-dive.md
```diff
@@ -19,11 +19,11 @@ Luckily with desmos the heavy lifting is done for us.
 If you haven't already done so, click [here](https://www.desmos.com/calculator/ouestfova3) for an example.
 You should see something as follows
 
-![Area between curves]({{ site.baseurl }}/assets/images/areaundercurve.PNG)
+![The image shows the correlation between how much time a person can loose when choosing to invest in continuous learning. The first line represents 20% investment of time for a 10% trade off. The second line represents 30% investment of time for a 20% trade off.]({{ site.baseurl }}/assets/images/areaundercurve.PNG)
```
RachelMuston, line 23: I would describe this image a bit differently.....e.g. this image is a 2D line chart with an x axis going from 0 to 8 and a y axis going from 0 to 8. The there are two lines on the chart; a red line and a blue line. Both start at point 0,0. The blue line is straight and ends at point x=8, y=8. The red line is curved and ends at x=~7.75, y=8. The two lines intersect at point 6.192. The area under the curve (between points 0,0 and the intersection at point 6.192) is coloured in green.

RachelMuston: Or something like that. :-)

_posts/en/2019-10-15-case-continuous-improvement-deeper-dive.md
```diff
@@ -19,11 +19,11 @@ Luckily with desmos the heavy lifting is done for us.
...
 
 What are we seeing here? If you hover over where the two lines intercept, you will see that the x value is 6.192, which is the same 6.192 you will see in the integral
 
-![Area under curve function]({{ site.baseurl }}/assets/images/integralareaundercurve.PNG)
+![The image shows an integral which displays a number, ~18.075. This is the calculation of the area under the curve f(x) between two lines that represent time investment in continuous learning, x=0 and x=6.192. ]({{ site.baseurl }}/assets/images/integralareaundercurve.PNG)
```
RachelMuston, line 27: Instead of having an image here...what if we wrote this in markdown instead? I've just been looking up 'Math equations in markdown' and I think we could write this formula as follows (but I haven't tested this):  $$\int_0^6.192 f(x) \,dx=18.0750720063.  And then we could eliminate this image and the need for alt text altogether.  Here is a site with more info (scroll down for examples):  https://math.meta.stackexchange.com/questions/4497/how-to-show-the-integral-symbol-on-this-site/

RachelMuston (changes requested):

smellems:
Closes #730

RachelMuston (approved):

PR #1391, opened 2020-12-01
Reduce width of webpages (excl. tables) + Adjust font sizes

gabrielcossette:
Hi,

Here's an attempt to reduce the width of our webpages to make them more readable. They should now look similar to articles on [Medium.com](https://medium.com) for example.

One exception is the tables. I've found that making them more narrow would hurt readability, so I left them wide (but they'll auto resize according to their content).

Finally, I've also adjusted the font sizes to make them more readable (headers, text...)

I'm not an UI/UX specialist so let me know if we need to tweak these changes. Also, I'm sure there are more elegant ways to code this so I'm open to any suggestions :)

PR #1392, opened 2020-12-01
Reduce width of blog posts (excl. tables) & adjust font sizes

gabrielcossette:
Similar to #1391 but only for blog posts for now.


RachelMuston (commented):
I like this! Makes sense to just make the change on the blog posts (for now).  If I was to make one other change I would make the blog title font bigger as there doesn't seem to be enough size difference between the title text size and the body text size. That being said, I would be ready to approve this PR if others are also on board.:)

gcharest:
The narrower setup for the blog posts makes sense but Medium usually fills the sides  with extra details/tools. So it's not extra white space for the sake of white space as much as adding value to it as far as I understand:
image

gcharest (commented):
As noted in my previous comment, I feel like there's something missing with all the white space around the text. We probably need to work out the overall elements' positioning in the available space. Usually pretty easily done with Bootstrap and responsive design so maybe we could take the time to put some wireframes together?

@ElminaIusifova what do you think?

ElminaIusifova (changes requested):
I might agree with the font size 20px ( and this can probably apply to all our website content) but not with the width.  Here is my advice:
Website content should be a minimum of 16px.  (What is our current one?)  18px is a great size for all our websites including blogs. Personally, I find 20px is a bit too big. 
Then, the H2 headline size is 38px, the H3 one is 25px, while the body text is 18px.
As for width: I would really leave the blog width aligned with the first header. In addition, scrolling takes too long, especially for long blog posts.

gcharest:
> As for width: I would really leave the blog width aligned with the first header. In addition, scrolling takes too long, especially for long blog posts.

You mean that by making the width narrower the scrolling gets too long? Should that be part of the work to break down blog posts in series instead? (Just trying to understand best approach)

Also thinking about length of blog posts. Medium has an indicator of the reading time based on length. Might be an easy way to calculate that based on number of words. Apparently the average reading speed is 236 words per minute according to this summary of a study: https://digest.bps.org.uk/2019/06/13/most-comprehensive-review-to-date-suggests-the-average-persons-reading-speed-is-slower-than-commonly-thought/

ElminaIusifova:
By the way, yes ;) some blogs are too long and we will have to break them down into two parts.  The ideal length of a blog post is between 6-7 minutes read, approx. 1,600 words.

gabrielcossette:
I updated the font sizes to the following (values are not exact because I'm using the scalable "rem" CSS unit):

&nbsp;&nbsp;&nbsp;H1 = 44,96px
&nbsp;&nbsp;&nbsp;H2 = 37,92px
&nbsp;&nbsp;&nbsp;H3 = 24,96px
&nbsp;&nbsp;&nbsp;H4 = 20px
&nbsp;&nbsp;&nbsp;Body = 17,92px
&nbsp;&nbsp;&nbsp;Date, Version… = 16px

RachelMuston:
Hey @gabrielcossette, when I look at the micro-acquisition status update blog in this branch, I find there isn't enough distinction between the title font size and the h2 font size.  Is there a way to make the title font size bigger? Or the font sizes for the subsequent headings smaller?

RachelMuston:
But maybe it's just me?  This is what it looks like: 
Screen Shot 2021-01-06 at 9 52 59 AM

gabrielcossette:
PR updated so that Title header of a blog post is now H1 (instead of H2).

ElminaIusifova (approved):

PR #1402, opened 2020-12-04
Merge more DORA questions

smellems:
Take questions from the DORA self assessment and merge with our questions.

I added a generic : Please rate how strongly you agree or disagree with the following statements. On my team: ..

and removed mention of my team in each statement.


gcharest (approved):
I like the new way of writing the questions! :+1:

PR #1403, opened 2020-12-06
Micro-Acquisition branding and content

RachelMuston:
Added a new Micro-Acquisition page using the one pager text
Removed the old Micro-procurement page (archived the content in a word file on the Sharepoint site)


gcharest (changes requested):
I'm OK with the rebranding but we probably should include a redirect if we're changing the pages URL?

Will double check if we did that in the past. If not I've done it on my site (gcharest.ca)

RachelMuston:
Oh geez, good point @gcharest!  We definitely should have a redirect on the old micro-procurement page.

RachelMuston:
we don't have a page-redirect.md file in the includes folder on our site so I'm assuming we haven't done before. I'll create a separate PR to add the includes file. :)

gcharest:
I'm using the jekyll-page-redirect plugin on my site. Might be useful here but will need to test out

gcharest:
Plugin: https://github.com/jekyll/jekyll-redirect-from

RachelMuston:
Ah interesting....shifting comments on this over to the new PR for the redirect.... #1405

gcharest (approved):
Approved with latest changes!

gabrielcossette (approved):
Great work! It's good to go :)

PR #1405, opened 2020-12-07
Add page redirect functionality to site

RachelMuston:
these changes required in order to add a redirect on the old Micro-procurement page (to point to the new Micro-Acquisition page).


RachelMuston:
Have added the code to the config.yml file but this gem install needs to be run (jekyll-redirect-from) 
 Not sure where to put this code.... more here: https://github.com/jekyll/jekyll-redirect-from

gcharest (commented):
OK, will be pushing a quick change shortly :)

gcharest:
Works perfectly, will need to update the page with a new front matter:

`redirect_to: <permalink to redirect to>`

gcharest (approved):

PR #1425, opened 2020-12-14
Fix webpage anchor scrolling

gabrielcossette:
Reimplement PR #819 (don't know when/why it was undone).

Original issue:

> Because we have a fixed navbar, linking to an HTML page anchor scrolls too far down (anchor is hidden by navbar).
> 
> With this fix, we can link to headers (1-6) and scrolling works as expected. And the new CSS doesn't seem to change the visual aspect of headers.
> 
> Open to a better workaround of course :)


gabrielcossette:
Sorry, I had an incorrect CSS wrongly added. It's all good now :)

gcharest:
I think I removed it at some point by mistake (looked at file history and saw I had removed the lines)

Not sure why, think part was because Jumbotron was out of place when in iPad screen size (still is in this setup by the way :yum: )

image

gcharest (changes requested):
Would need to look at the iPad screen size. Seems to be an issue (see previous comment)

gabrielcossette:
@gcharest Good catch! The menu items wrap because of the reduced size.

We could make our menu items less verbose. I renamed:

Our Documents | Our Blog Posts
to
Documents | Blog

What do you think?

smellems (approved):

gcharest (approved):

gabrielcossette:
I was able to fix the menu overflow when text wrap (e.g. in tablet mode).

PR #1442, opened 2021-01-07
Convert from Travis CI to GitHub Actions

gabrielcossette:
This PR converts our repository from Travis CI to GitHub Actions.

GitHub Actions are more integrated in GitHub. However, one caveat is that they don't support colouring of all outputs yet. This means that the errors from link-check can't be in colour for example.
To work around this, I disabled the "Reading:" and "Warning:" lines display from link-check. Unless I'm not mistaken, I don't think they were really useful anyway.

Let me know if something should be done differently :)


gabrielcossette:
Sorry, I had forgotten to disable the link-check warnings. All good now :)

gcharest:
Overall, I don't see this as an issue (coloring) but I do have a question:
Wouldn't more integrated with GitHub mean less interchangeable when thinking of hosting options? We'd need to redo this if we were to change service provider for example?

gcharest (approved):
Seems to work fine here!

gabrielcossette:
> Overall, I don't see this as an issue (coloring) but I do have a question:
> Wouldn't more integrated with GitHub mean less interchangeable when thinking of hosting options? We'd need to redo this if we were to change service provider for example?

That's a good question and a very valid concern. I'm unfortunately not knowledgeable enough about CIs to respond. But as our GitHub Actions script is pretty basic, I would say there are more benefits for us in all cases, at least for now.

PR #1462, opened 2021-01-14
Change MathJax CDN to jsdelivr

### LINKED ISSUE #1401: cdn.mathjax.org retired - Error loading script
While navigating the team,s site, I noticed an error in the console:

>MathJax.js?config=TeX-MML-AM_CHTML:32 WARNING: cdn.mathjax.org has been retired. Check https://www.mathjax.org/cdn-shutting-down/ for migration tips.

Led me to quickly lookup why this was popping up on all pages: turns out we load the MathJax script in all pages. Should probably only enabled when there's a need for it to make site more lightweight but this separate issue.

Right now, the script is apparently loading from mathjax.org and that causes an issue.

Option could be to load from jsdelivr.com (https://www.jsdelivr.com/package/npm/mathjax) or other CDN.

smellems:
Closes #1401 

Quick fix to remove warning about mathjax cdn being retired (since 2017).

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -34,7 +34,7 @@
     href="https://cdnjs.cloudflare.com/ajax/libs/datatables/1.10.19/css/dataTables.bootstrap4.min.css">
 
   <!-- Mathjax Support -->
-  <script type="text/javascript" id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js">
+  <script type="text/javascript" id="MathJax-script" async src="https://cdn.jsdelivr.net/npm/mathjax@2/MathJax.js?config=TeX-MML-AM_CHTML">
   </script>
 
-  </head>
+</head>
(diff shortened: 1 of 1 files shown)


gcharest:
I've been scouring through the whole set of blogs/pages... we don't call MathJax anywhere :thinking: 

Might just pull it out altogether?

gcharest:
Propose we remove it entirely

gcharest (changes requested):
Need to confirm if even used anywhere in the site - Otherwise propose to remove script entirely.

gcharest (approved):

gcharest (changes requested):
cdn doesn't work, mathjax used in the Case for Continuous Improvement - Deeper Dive post.

Right under the first graph:

image

smellems:
Fixed by using latest version 2 instead of 3.

Upgrading to version 3 requires some changes.  Version 2 is still supported, for now.
http://docs.mathjax.org/en/latest/upgrading/v2.html

> MathJax v3 is still a work in progress; not all features of version 2 have been converted to version 3 yet, and some may not be. MathJax v2 will continue to be maintained as we work to move more features into version 3, but MathJax v2 likely will not see much further development, just maintenance, once MathJax v3 is fully converted.

gcharest (approved):
Works perfectly :+1:

PR #1511, opened 2021-01-28
Website changes - files, pages and menus

smellems:
This is to implement the changes discussed by the team to clean up the Website.

- All pages that have moved have a redirect from the old URL.
- I added sub-folders to pages for archives and redirects to help organize files in our repo, this does not affect the URL (since we use permalink)
- Medium and Long term IT pictures have been merged into "IT Vision".
- Lots of pages were archived.
- Documents and presentations split into 2 pages and documents renamed to strategies.
- Team page is now about us page.
- Add Initiative page for "work in progress" section
- Add status for strategies and initiatives

Still to do
- Add contact us links and emails..
- Create french versions of missing pages with notice (need to translate)

PR #1531, opened 2021-02-05
Switch to Bootstrap Sass & Render mobile view earlier

gabrielcossette:
Upcoming PR #1511 makes the menu larger, so we need to render the mobile view earlier so that the menu doesn't overflow (it was affecting tablets in portrait mode).

This PR switches Bootstrap CSS to dynamic generation through Sass (incl. minify), to which we add our custom breakpoint variables.

After merge, to render the website locally, we'll have to run `bundle install` once to install Gem dependencies, and then use the command `bundle exec jekyll serve`. 

Let me know if you have any questions :)


smellems:
Is there a way we could use the Bootstrap gem, instead of adding all the files in our project?
https://rubygems.org/gems/bootstrap/

Also we could look at upgrading Bootstrap at the same time.  4.6 is the latest.

gcharest (changes requested):
Looks good but I'm having an issue when the screen size goes below a certain size. I guess there probably should be a `min-width` set?

image

gabrielcossette:
@smellems Good idea! This PR now uses Bootstrap gem. It needs to be merged after our switch to GitHub Actions for website deployment (#1535). I'll update the documentation as team members will need to use Bundler commands from now on to test (e.g., `bundle install`, `bundle exec jekyll serve`)
Regarding updating Bootstrap, I'll start a new PR :)

@gcharest As it's related to "single words per line" overflowing, the solution I see would be to reduce the font size for this small screen width. However, as this is happening at around 230px width, are there devices of this small of a size?

.github/workflows/github-pages.yml
```diff
@@ -0,0 +1,15 @@
+name: Build and deploy Jekyll site to GitHub Pages
```
smellems, line 1: Do we need to move to Github actions?

I think that even without this file, the site would still be deployed in the same way it currently is

smellems:
This is nice!

I'm still using `docker run -p 4000:4000 -v $(pwd):/srv/jekyll -it --rm jekyll/jekyll:3 jekyll serve`
and this also fixes an issue, where I was having to install jekyll-redirect-from every time

.github/workflows/github-pages.yml
gabrielcossette: There may be another way using the regular GitHub Jekyll build, but I haven't found one yet.

The issue is that we need Bootstrap sources to build the CSS, and I don't know how to dynamically import it with the regular method (other than copying it in our repo).

Problematic line is `@import "bootstrap";` in `assets/css/style.scss`

GitHub error msg:
> Your site is having problems building: Your SCSS file assets/css/style.scss has an error on line 2: File to import not found or unreadable: bootstrap. Load paths: assets/css /hoosegow/.bundle/ruby/2.7.0/gems/jekyll-theme-primer-0.5.4/_sass. For more information, see https://docs.github.com/github/working-with-github-pages/troubleshooting-jekyll-build-errors-for-github-pages-sites#invalid-sass-or-scss.

But GitHub Actions does give us additional benefits:

> **Control over gemset**
> * **Jekyll version** — Instead of using the currently enabled version at `3.9.0`, you can use any version of Jekyll you want. For example `4.2.0`, or point directly to the repository.
> * **Plugins** — You can use any Jekyll plugins irrespective of them being on the [supported versions](https://pages.github.com/versions/) list, even `*.rb` files placed in the `_plugins` directory of your site.
> * **Themes** — While using a custom theme is possible without Actions, it is now simpler.
>
> **Workflow Management**
> * **Customization** — By creating a workflow file to run Actions, you can specify custom build steps, use environment variables.
> * **Logging** — The build log is visible and can be tweaked to be verbose, so it is much easier to debug errors using Actions.

gcharest (approved):

smellems (approved):

PR #1560, opened 2021-02-12
Add blog post navigation (previous, next)

### LINKED ISSUE #1465: Link blog posts series together
Simply make it easy for user to rotate through blog posts in a series.

Example: https://engineering.chrobinson.com/how-to/linking-a-series-of-jekyll-posts/

smellems:
Adds buttons to blogs, to navigate to previous and next blog post.

Example
image

Closes #1465

### DIFF
--- a/_data/i18n/general.yml
+++ b/_data/i18n/general.yml
@@ -82,6 +82,12 @@ blogs:
   translationDate:
     en: Translated on
     fr: Traduit le
+  previous:
+    en: Previous
+    fr: Précédent
+  next:
+    en: Next
+    fr: Prochain
 
 notices:
   notice:
--- a/_data/i18n/general.yml
+++ b/_data/i18n/general.yml
@@ -83,11 +83,11 @@ blogs:
     en: Translated on
     fr: Traduit le
   previous:
-    en: Previous
-    fr: Précédent
+    en: Previous blog
+    fr: Blogue précédent
   next:
-    en: Next
-    fr: Prochain
+    en: Next blog
+    fr: Prochain blogue
 
 notices:
   notice:
--- a/_includes/blog-nav.md
+++ b/_includes/blog-nav.md
@@ -2,22 +2,23 @@
 <div class="border border-dark" style="display: flex; justify-content: space-between; flex-wrap: wrap;">
   {%- assign posts=site.posts | where:"lang", page.lang -%}
   {%- for post in posts -%}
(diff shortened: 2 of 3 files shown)


gcharest:
Two things to consider:
- Navigation at the bottom only makes sense if we read the whole post but I'm wondering if it wouldn't be practical to have it at the top as well?
- The links should be equally spaced IMHO. Depending on the length of post titles, it makes it a bit inconsistent and sometimes the links run on top of each other. This [blog post explains](https://www.bytedude.com/jekyll-previous-and-next-posts/ */) how it's possible to use flexbox to better align the buttons on each side, even on mobile form.

gcharest (changes requested):
Equally spread both links on each side of the page.
Consider adding the posts navigation links at the top as well

gcharest:
Other reference in Bootsrap for navs and flex so that the buttons are equally spread: https://getbootstrap.com/docs/4.0/components/navs/#working-with-flex-utilities

smellems:
Updated version uses Flexbox with a border

image

Small screen

image

ElminaIusifova (changes requested):
If the purpose of this PR to navigate between blogs, the best way to add ARROWS (Arrow left -> "Previous", arrow right -> "Next"). I wouldn't put the name of the blogs on the button

ElminaIusifova:
If you still want to leave the name of the blogs on the buttons I would remove the stroke on the box, make the font colour of the content  different, and increase the margin-top size

gcharest:
There seems to be an odd behaviour when navigating around the posts :thinking: Like, some sort of loop?

gcharest:
Nevermind the loop, propably a cache issue. Looks great!

Propose we add a visual indicator for hover/focus so that users who see can have a cue of the selectable area. Tab navigation works fine.

gcharest:
> If the purpose of this PR to navigate between blogs, the best way to add ARROWS (Arrow left -> "Previous", arrow right -> "Next"). I wouldn't put the name of the blogs on the button

This change request has been actioned in [e031454](https://github.com/sara-sabr/ITStrategy/pull/1560/commits/e031454c1762ea68a042900b8eae64bee32ba320)

ElminaIusifova (changes requested):
Can we remove the box around and add the word "blog"? :
<- Previous blog     -> Next blog

ElminaIusifova (approved):

PR #1585, opened 2021-02-25
Website width justification

ElminaIusifova:


ElminaIusifova:
Justified text on websites is a bad idea :wink: 
Here is a resource about it   https://www.castlegateit.co.uk/2013/09/can-justified-text-be-justifed-for-the-web/.

PR #1616, opened 2021-03-17
Mandate page

ElminaIusifova:


_pages/en/mandate.md
```diff
@@ -8,40 +8,52 @@ sections: Mission
...
 
-## Mandate (team's authority within IITB)
 
-- Diagnose key organizational problems, waste, lead times and governance structures that inhibit the desired changes in behaviours
-- Recommend specific course of actions to governance bodies to adopt new methods of managing IT risks
-- Measure progress of adoption of strategies in IT teams and governance bodies as a performance indicator of the IT Strategy team
-- Challenge current thinking and approaches
 
-## Team's Guiding Principles
+### **Team's Guiding Principles**
```
RachelMuston, line 40: There is some overlap between this content on guiding principles and the content above on 'Our Values'. Can we condense into one section?

RachelMuston: Also wondering if this content (team values/principles) would be better on the About us page rather than the Mandate page? Just a thought.

_pages/en/mandate.md
```diff
@@ -8,40 +8,52 @@ sections: Mission
...
 Learning is a core part of daily activities to ensure up to date, modern, and well-researched direction.
 
-## Lead by Example
+- Lead by Example
 
 By using the approaches and technologies we champion, we are able to better respond and anticipate potential challenges.
 We also experiment with modern ways of working in order to demonstrate their value and effectiveness.
 
-## Team understanding of the "digital transformation"
+### **Team understanding of the "digital transformation"**
```
RachelMuston, line 56: Wondering about the rationale for this section? It doesn't seem to fit logically under 'mandate'. And it feels too specific to fit under 'About Us'. Would this content be better as a blog post?

_pages/en/mandate.md
```diff
@@ -8,40 +8,52 @@ sections: Mission
...
 ---
 
-## Vision (image of the future)
+- Diagnose key organizational problems, waste, lead times and governance structures that inhibit the desired changes in behaviours
+- Recommend specific course of actions to governance bodies to adopt new methods of managing IT risks
+- Measure progress of adoption of strategies in IT teams and governance bodies as a performance indicator of the IT Strategy team
+- Challenge current thinking and approaches
 
-The [Canadian Experience](human-development-life-cycle.html) is when Canadians are benefiting seamlessly of the federal government services through modern technologies.
+<img src="assets/images/our-values.png" alt="">
```
RachelMuston, line 16: Buzz kill alert.....I'm pretty sure that using images instead of headers breaks accessibility rules (even with alt text). :-(

RachelMuston (commented):
Left a couple of comments. :-) In general I think this page is great. The content is clear and easy to read.

_pages/en/mandate.md
```diff
@@ -8,40 +8,52 @@ sections: Mission
...
+- Learning and Continuous Improvement
+- Self-Discipline
```
RachelMuston, line 18: I can't put it into words well but 'Self-Discipline' as a value leaves me with an unpleasant feeling. It feels...judgemental.  Curious what the thinking was behind that one.

_pages/en/mandate.md
```diff
@@ -8,40 +8,52 @@ sections: Mission
...
+- Innovation
+- Collaboration
+- Communication
+- Teamwork
```
RachelMuston, line 23: Also curious why 'Working in the open' isn't on our values list?  I'm thinking about...when a new team member joins our team, what would we tell them are the values of the team?

_pages/en/mandate.md
RachelMuston: What I really like about the principles is that we explain what the principle means for our team. If I was to combine the values list and the principles list, I would keep the format and content of the principles list and add in:

"Teamwork - we swarm to solve problems and respond to requests, we have our own files but we also work across files sharing our expertise to help one another" or something like that and 

"Working in the open - we strongly believe in sharing our work openly on this website. We include here not just final products but also products in development so that we can collaborate and improve our work and also so that our work can be re-used by anyone." or something like that.

_pages/en/mandate.md
RachelMuston: Ok it seems like it might be possible to make image headers accessible but I'm not sure if this solution meets GC accessibility requirements.... https://www.webaxe.org/accessible-headers-with-graphics/

_pages/en/mandate.md
ElminaIusifova: This section originally was on the mandate page. I like the idea of making it a blog. Should we remove it from the Mandate page?

_pages/en/mandate.md
ElminaIusifova: Might not be the best word but I mean "we take responsibility for managing our workload and  participate in team decision making to come up with a solution to any problem statement"

_pages/en/mandate.md
ElminaIusifova: Missed this one. Working in the open is one of our values.

_pages/en/mandate.md
ElminaIusifova: ok, I will leave the format of "Team principles" and will add those two points. Thanks ;)

_pages/en/mandate.md
ElminaIusifova: As for images, it was just a try. If it breaks accessibility and GC rule I will return simple headers in markdown.  Just wanted to see the team's opinion on it. 
There is a way to do Graphic headers in HTML accessible not sure about markdown. I will wait for what the rest of the team will say otherwise will return H1.

smellems (approved):

gcharest (approved):

PR #1738, opened 2021-06-03
Stats on digital policy instrument suite

remyb2canada:
Blog to publish some statistics around the number of requirements that some senior officials are expected to be accountable for. Especially the CIO.

Puts emphasis on the role of the CIO in bringing digital to a reality.
That LOBs need to develop stronger technical acumen, and that IT staff need to foster stronger empathy towards their business counter parts (e.g. it's not always about code, it's about policy and preparing for changes coming up from those policy changes).

**NOTE 1**: footer.html changed _includes/footer.html to use an absolute reference to datatable.js (using site.baseurl) as I got errors from *.md files in a posts folder

**NOTE 2**: french translation will be pushed after peer review (and before merge)


gcharest (approved):
I've made a few tweaks to spelling. Overall blog is both well written and a bit scary. 1 ADM at a table of 20+ ADMs responsible for successfully implementing 55% of the policy requirements. Power struggles must be terrible.

PR #1741, opened 2021-06-07
Translate presentation to French - small changes

smellems:


gcharest (approved):
Approved but will need to fix the image on the slide "How it works"

In the upper right hand, the sentence spells doing with two "i": "Your team is **_doiing_** well"

Issue #1629, opened 2021-03-30
Resources page changes


ElminaIusifova:
- Remove the ‘Some of that content may only be available on GC/ESDC network so if the link doesn’t work externally, we apologize for the inconvenience.’ sentence.  Instead next to each link that is only ESDC network add ‘ESDC network’.
- Also if a link is only in English add ‘english only’ and  ‘anglais seulement’ to the french page.  This is a best practice.

PR #1778, opened 2021-06-30
Resources page changes

ElminaIusifova:
closing #1629 
Resources page:
@smellems   can you please add to each link that is only ESDC network ‘ESDC network’ label.
and if a link is only in English add ‘english only’ and ‘anglais seulement’ to the french page just like you did with strategies and initiatives pages.

### DIFF
--- a/_data/tables/references.yml
+++ b/_data/tables/references.yml
@@ -13,19 +13,19 @@
             items:
               - item:
                   fr: |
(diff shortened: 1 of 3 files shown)


gcharest:
Oh, that's a lot of changes :thinking: Will get started in between other tasks

gcharest (approved):
I think this covers most of the elements you mentioned @ElminaIusifova :+1:

PR #1780, opened 2021-07-07
Update issue templates - adding MA template

RachelMuston:
I created a template last week for MA issues but the template isn't showing up in the 'New Issues' page (but it can be found by clicking on 'Edit Templates'). Hoping this PR adds it to the New Issues page (this time I created it via settings directly on the New Issues page). FYI I don't know what I'm doing. lol


gcharest (approved):

gcharest (changes requested):
I think the about text was not updated :smile:

gcharest (approved):

PR #1811, opened 2021-07-22
Product mgt defs

remyb2canada:
In strategies / supporting documents.

An inventory of definitions for the word "product". And what ESDC is to chose. Will be updated as this journey continues...


gcharest:
Love the growing list! Also shows how much perspectives exist on the same term... :thinking:

gcharest (approved):

PR #1852, opened 2021-08-11
improved pandoc instructions

RachelMuston:


ElminaIusifova:
Now it is very clear with the names of the file.  
As for the directory: do you have to make sure you are in the same repo or in the same folder where the markdown file was copied?

ElminaIusifova (approved):

PR #2142, opened 2022-04-27
Closeout blog - merge on May 5

RachelMuston:
For review :)


RachelMuston:
> * **"The five things required in order to be successful with a coding micro-procurement initiative in the Government of Canada".**
>   Can you change the font size and make it smaller than the title? maybe just make it bold?

Yeah I don't love it either but its a formatting that defined in the CSS (for headings ). Changing this will change the formatting across the site. Although I have always felt that the size of the H2 text was always too close to the size of the H1 text....

remyb2canada (approved):

PR #2582, opened 2023-12-05
Fixes to dead links

RachelMuston:


RachelMuston:
fyi I tested increasing the timeout to a longer time to see if that helped (it did not), so I reduced the timeout time again.

PR #2584, opened 2023-12-07
Quiet link check

RachelMuston:
Testing to see if I can get the link checker to show the dead links but not prevent publishing.


RachelMuston:
No longer needed as Steven has put together a much better fix.

PR #2588, opened 2023-12-12
Aded workflow to check dead links

RachelMuston:
Hey @smellems will this work? I'm trying to set up an action to run weekly to check for deadlinks on the website.

I couldn't test it locally - scheduled actions can't be tested in forked repos or local branches.


smellems (approved):
Let's try it!

PR #2667, opened 2024-05-10
Update package.json and package-lock.json so test script warns on broken links

thebiggest0:
Package.json
- Change devDependencies for mark-down-link from “^3..10.3” (https://github.com/tcort/markdown-link-check) to https://github.com/thebiggest0/markdown-link-check.
- Reason: To prevent an error from being thrown when a broken link is encountered, the javascript in markdown-link-check needed to be changed. And for the change to be retained for future use this was the solution I found.

Package-lock.json
- Update content to reflect changes made to package.json (through npm install).
- Changes: mark-down-link resolved points to https://github.com/thebiggest0/markdown-link-check instead of the original registry on http://npmjs.org/.
- Changes: The version is now 3.10.4 (before 3.10.3) to reflect the changes I made in the markdown-link-check file. In the function runMarkdownLinkCheck() in the if condition which checks for broken links I changed console.error to console.warn and removed resolve(), now promise does not fail when encountering a broken link.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "editorconfig-checker": "^4.0.2",
         "license-compatibility-checker": "^0.3.0",
         "licensee": "^10.0.0",
-        "markdown-link-check": "^3.10.3",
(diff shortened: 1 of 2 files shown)


RachelMuston (approved):
Nice work! :)

PR #2668, opened 2024-05-14
Update broken links in posts

thebiggest0:
Remove broken links in _posts folder
- en 2020-08-18-better-tech-through-microprocurement-p1
- fr 2020-08-18-meilleure-technologie-grace-au-micro-approvisionnement-partie-1


RachelMuston (commented):
Small fix needed - the english and french need to match exactly, also we've found that it can be useful to keep the full link for reference purposes). To ensure that the link-checker doesn't pick it up as a link, I think we removed the hyperlink markdown []() and removed the 'https://www' from the link. So the fix for line 110 in the english file would be 'The data (ic.gc.ca/eic/site/061.nsf/vwapj/SFGSME-EFCPME_2017_eng_public.pdf/$file/SFGSME-EFCPME_2017_eng_public.pdf) from the...'

RachelMuston (commented):
Well done on adding the unknown words to cspell. It looks like you added some words twice - SFGSME and EFCPME. You should only need to add each word once and then it will be recognized every time it appears.

RachelMuston (approved):

